Issue #4, opened 2020-04-17
Fix /party 404 issue


morganecf:
This happens when you try to load the /party page or /about or /map or whatever without visiting the front page first

PR #17, opened 2020-04-21
Serve react from flask

morganecf:
This addresses issue https://github.com/morganecf/jitsi-party/issues/4.

Right now we use react routing to navigate between pages. This is just client-side routing, meaning that the server isn't hit when we change URLs. The server does get hit when we _refresh_ after changing URLs, and it'll return a 404 since we don't do any server-side routing.

Afaik the 3 options to address this are:
1) Ditch react routing altogether and just have an SPA without any URL changes 
2) Use hashHistory instead of browserHistory to simulate going to different URLs (they'd look like `/#/party`)
3) Have some form of server-side routing

This PR implements 3) and just does the simplest thing, which is to redirect to the welcome page if the user tries to refresh. It might be easier to do 1), which would have the same UX effect as this PR, if adding flask right now presents deploy complications.


SeanMcGrath:
FWIW i'm in favor of option 1 or 2, unless we have other plans for the backend - in which case, might as well do this now

PR #18, opened 2020-04-21
More css/styling improvements

SeanMcGrath:
makes the buttons a bit nicer, reclaims some space for the nav. makes some stuff fixed to the screen for a more consistent exp


app/client/jsx/routes.jsx
```diff
@@ -2,7 +2,6 @@ import React from 'react'
 import { Router, Route } from 'react-router'
 import { createBrowserHistory } from 'history'
 import Welcome from './Welcome.jsx'
-import About from './About.jsx'
```
SeanMcGrath, line 5: this jsx is gone, had to do this to fix build

PR #39, opened 2020-04-29
Add muteRoom option to rooms.

seangibat:


app/client/jsx/JitsiVideo.jsx
```diff
@@ -18,10 +18,17 @@ class JitsiVideo extends Component {
...
         // These should not be stored in state for the duration of the Jitsi video,
         // since any change to state currently triggers a render and reconnect
         this.isAudioMuted = this.props.isAudioMuted
         this.isVideoMuted = this.props.isVideoMuted
+
+        // If this is a muteRoom, mute and remove microphone controls.
+        if (this.props.jitsiData.muteRoom) {
+            _.pull(this.toolbarButtons, 'microphone')
+            this.isAudioMuted = true;
+        }
```
seangibat, line 31: While I could've done `this.isAudioMuted = this.props.jitsiData.muteRoom || this.props.isAudioMuted`, I thought it would be better to put all the `muteRoom` options in one place.

app/client/jsx/Room.jsx
```diff
@@ -39,7 +39,8 @@ class Room extends Component {
            displayName: this.props.displayName,
            avatar: this.props.avatar,
            roomName: roomData.name,
-           height: roomData.videoHeight
+           height: roomData.videoHeight,
```
morganecf, line 42: Is this still used or do we rely on room-specific css for this?

morganecf: Also asking bc I'm wondering whether or not to put it in the db schema

seangibat: It's still used in jitsi rooms, looks to me.

seangibat: Or all rooms with Jitsi? It's in the Jitsi class, but it seems like video expands to the space?
```
            const options = {
                roomName: this.props.jitsiData.roomName,
                height: this.props.jitsiData.height || 500,
                parentNode: document.getElementById('jitsi-container'),
                interfaceConfigOverwrite: {
                    // filmStripOnly: true,
                    SHOW_JITSI_WATERMARK: false,
                    DEFAULT_REMOTE_DISPLAY_NAME: 'Fellow Clarendonite',
                    SHOW_WATERMARK_FOR_GUESTS: false,
                    TOOLBAR_BUTTONS: this.toolbarButtons,

                },
                configOverwrite: {
                    disableSimulcast: false
                }
```

seangibat: Yeah it's not used anymore, CSS is overriding

morganecf: Mmk, would you be able to get rid of those references in this PR? And I won't include it in the db models

seangibat: Sure

PR #40, opened 2020-04-29
Add muteRoom to Dance Party; fix muteRoom.

seangibat:


hndrewaall:
Why is there a capacity to the dance party at all? Shouldn’t it be unlimited?

hndrewaall:
test

seangibat:
idk, to make it a scarcer resource or something? i can remove it

hndrewaall:
i think we should remove it, but i’ll let @morganecf chime in

morganecf:
@hndrewaall the capacities are completely inactive right now. I just added them to the room config when I first made the config and never removed them, but I'll clean all that up when I work on enforcing room capacities for certain rooms like closets/bathrooms.

app/client/jsx/JitsiVideo.jsx
```diff
@@ -73,7 +73,7 @@ class JitsiVideo extends Component {
                     SHOW_JITSI_WATERMARK: false,
                     DEFAULT_REMOTE_DISPLAY_NAME: 'Fellow Clarendonite',
                     SHOW_WATERMARK_FOR_GUESTS: false,
-                    TOOLBAR_BUTTONS: this.toolbarButtons,
+                    TOOLBAR_BUTTONS: toolbarButtons,
```
morganecf, line 76: did you remove this in previous PR? oops

seangibat: hm no, I think you may be readin the PR wrong

seangibat: hm no, I think you may be readin the PR wrong

PR #43, opened 2020-04-29
Backend

morganecf:
This isn't ready to merge yet but feedback is welcome.

The problem I'm trying to solve here is how to mark users as online/offline, so that the client can obtain a faithful state of who's in which room. The way we do this is:

1. User leaves/enters room: send socket update, server responds with broadcast to all clients with room that needs updating, clients fetch user list for that room, endpoint only returns active users
2. User closes tab or refreshes: `window.beforeUnload` event captures this and does 1)
3. Unexpected client crash/disconnect: flask-socketio broadcasts disconnect event if hasn't received a heartbeat in x seconds. x should be configurable but I have not touched it. 

THIS ALSO IMPLEMENTS DOORS! Kinda. There isn't any "close door" functionality yet, but there is now a step. You can see all the socket stuff working 

This also moves the room layout into `rooms.json` and adventures into `adventures.json`. Rather than storing the potentially in-flux room definition in a db, I'm only storing room name/id/capacity. Room id is needed to update the other tables, but the rest of the room definition is really just used by the client, so I'm going to add a route to serve the json file without worrying about putting everything in the db. Running `flask create-db` populates the db from scratch with the rooms in `rooms.json.`

The db models are based on the schema Andrew came up with in the google doc.

To do:
- [x] Create endpoints to get active users for room / all rooms
- [x] Mark vestibule as visited
- [x] Basic door functionality
- [x] Event listeners for room enter/leave, refresh rooms
- [x] Consolidate room/adventure definition into one file, serve json

Going to do rabbitMQ in a separate PR. It isn't required for local dev.


hndrewaall:
I'm a little worried about spawning threads inside Flask workers.

app/jitsi/main.py
```diff
@@ -0,0 +1,33 @@
...
+
+    '''Start a background thread that marks stale users as offline'''
+    def find_offline_users(app):
+        with app.app_context():
+            while True:
+                update_stale_users()
+                time.sleep(10)
+
+    thread = Thread(target=find_offline_users, args=(current_app._get_current_object(),))
+    thread.start()
```
morganecf, line 20: @hndrewaall what do you think of this approach?

morganecf:
@hndrewaall could we have a script outside of flask that does this?

hndrewaall:
I think maybe the more correct approach would be async tasks. Something like you'd get with Celery

hndrewaall:
https://docs.celeryproject.org/en/latest/userguide/periodic-tasks.html

hndrewaall:
Yeah all the relevant state is in the DB right? Celery workers can use the same SQLAlchemy models and interact with the DB directly. Though at this point we might start looking at Postgres, as you have multiple processes playing with a single db (actually this is true even with just multiple web workers)

hndrewaall:
woooo complexit!

app/client/jsx/Room.jsx
```diff
@@ -24,6 +25,10 @@ class Room extends Component {
         this.state = {
             room: this.props.currentRoom
         }
+
+        this.service = new Service()
```
SeanMcGrath, line 29: FYI `service` is super generic - what kind of service?

app/jitsi/main.py
SeanMcGrath: I think you can do away with `user.online` and just check the value of `last_seen` at lookup time

SeanMcGrath: FYI I think the "correct" way to do this, if it turns out to be necessary, is to have a separate "db manager" application that runs completely offline, rather than a thread within this online app

app/client/jsx/Exit.jsx
```diff
@@ -1,17 +1,21 @@
...
 import { connect } from 'react-redux'
 import avatars from './avatars.jsx'
+import Service from './Service.jsx'
 
 class Exit extends Component {
     constructor(props) {
         super(props)
+
+        this.service = new Service()
+        this.service.disconnect()
```
SeanMcGrath, line 11: the fact that you have to init and then disconnect here is suspicious

seangibat:
Agreed that celery is ultimately the right approach, but might be some quick easy hacks in the meantime:
1) Threads seem fine to me short-term
2) Or trying using APScheduler https://apscheduler.readthedocs.io/en/latest/ (I've used it at work to great success)
3) Or check on all users' state when a user hits the socketio keepalive and its been a certain amount of time since the last check

hndrewaall:
Won't you potentially end up with tons of threads spawned all doing the same thing?

app/jitsi/database.py
```diff
@@ -0,0 +1,39 @@
+from . import db
+from datetime import datetime
+from .models import User, Room, UserRoomState, UserLocation
+
+#TODO make timeout a part of config, in seconds
+USER_TIMEOUT = 30
+
+def create_user(username, avatar):
```
hndrewaall, line 8: I feel like these should be class methods of the `User` model, no? I think we don't really need this module and can move this all there

seangibat:
> Won't you potentially end up with tons of threads spawned all doing the same thing?

Yeah but, that's probably fine, isn't it? I don't think it would overload the server or DB. Race conditions would be our biggest worry, but it _seems_ like we wouldn't really have to worry about that - the background check would just be running more often.

app/client/jsx/FlaskInterface.jsx
```diff
@@ -0,0 +1,82 @@
+import axios from 'axios'
```
hndrewaall, line 1: apologies for nitpick, but could we rename so as not to bind to a specific backend framework? could we call this WebApi, or something else generic?

hndrewaall: as in, not Flask

app/client/jsx/FlaskInterface.jsx
```diff
@@ -0,0 +1,82 @@
...
+import io from 'socket.io-client'
+import { createAdventureActions } from './utils.js'
+
+const url = 'http://127.0.0.1:5000'
+const PING_INTERVAL_MS = 10000
+
+
+export class Socket {
```
hndrewaall, line 9: more nitpick: these are both APIs, right? One just uses WebSockets as transport, the other uses HTTP? so why not call them WebSocketApi  and HttpApi or something?

hndrewaall: also, I've totally forgotten, but is there a reason to not have all our API methods over WS?

app/jitsi/database.py
```diff
@@ -0,0 +1,21 @@
+from . import db
```
hndrewaall, line 1: I still think this file is unnecessary and these methods can be `User` model class methods

hndrewaall: or Room

morganecf: This is the global sqlalchemy `db` object. I moved those other methods into the User class

app/jitsi/models.py
```diff
@@ -0,0 +1,119 @@
...
+    @staticmethod
+    def create(username, avatar):
+        # TODO change this once Daniel fixes avatars
+        avatar = '-'.join(map(str, avatar))
+        user = User(username=username, avatar=avatar)
+        db.session.add(user)
+        db.session.commit()
+        return user
+
+    @staticmethod
```
hndrewaall, line 33: these should probably be class methods so you aren't constantly typing the class name

morganecf: These don't operate on a user instance, though!

PR #48, opened 2020-05-01
Cabin Fever

danielmdavis:


app/client/jsx/PuckSelect.jsx
```diff
@@ -8,107 +8,99 @@ class PuckSelect extends Component {
...
+        {all_avatar_designs_array}
         </div>
       )
 
-    }
+    } else {
 
+      return ( // when avatar state isn't null, collapse all and display choice
+        <div className='outer'>
+          <input className='fade' type="button" onClick={this.onButtonClick} value="Pick your puck"/>
```
morganecf, line 97: Out of curiosity, why does `this.onButtonClick` not require a bind? It uses `this.setState` which makes me think it would, though I do get how it's working nonetheless

app/client/jsx/Welcome.jsx
```diff
@@ -36,27 +36,30 @@ class Welcome extends Component {
...
           border: '2px solid #D6D3CD',
+          backgroundColor: '#D6D3CD',
           borderRadius: '7px',
-        //   backgroundColor: '#222426',
-          color: '#D6D3CD',
+          color: '#1B1E1F',
           outline: 'none',
           boxShadow: 'none',
-          fontSize: '14px'
+          fontSize: '20px'
```
morganecf, line 44: I think we should definitely keep styles outside of JS

app/client/styles/abstracts/_variables.scss
```diff
@@ -2,13 +2,15 @@
 // This file contains all application-wide Sass variables.
 // -----------------------------------------------------------------------------
 
-
-
+@import url('https://fonts.googleapis.com/css2?family=Alegreya&display=swap');
+@import url('https://fonts.googleapis.com/css2?family=Darker+Grotesque:wght@600;800&display=swap');
```
morganecf, line 6: Could we download the fonts so we don't have to fetch them each time?

app/client/styles/components/_puckBox.scss
```diff
@@ -6,46 +6,50 @@
...
 	max-height: 70px !important;
 }
 .non-selected-image {
 	text-align: center;
-	min-width: 42px !important;
-	width: 42px !important;
-	margin: 8px 4px 0px 4px;
+	min-width: 38px !important;
+	width: 38px !important;
+	margin: 4px 4px 0px 4px;
```
morganecf, line 37: There are a lot of hardcoded values and !important in this code...I think we should figure out a way to clean this up without resorting to those things, but we can do this in a subsequent PR

app/client/styles/components/_Welcome.scss
```diff
@@ -2,6 +2,26 @@
...
+h1 {
+	margin: -20px 0px 30px 0px;
+	-webkit-text-stroke: 1.25px #1B1E1F;
+	font-weight: 800;
+	font-size: 2.25em;
+}
+
+input[type="button"] {
+	background-color: #D6D3CD;
+	color: #1B1E1F;
```
morganecf, line 23: Can these be sass variables? You reuse the colors a few times in this sheet, so we could abstract them at the top: $button-background-color: #D6D3CD

app/client/jsx/PuckSelect.jsx
```diff
@@ -8,107 +8,99 @@ class PuckSelect extends Component {
...
-              // if (id===0) { selected = 'box-selected' }
-              let handleClickColor = () => this.handleClickColor(id)
-
-              // avatar color variant selection
-              return (
+              let handleClick = () => this.handleClickColor(id)
+              return ( // returns all colorways for selected design
                 <PuckBox
-                handleClick={handleClickColor}
+                handleClick={handleClick}
```
morganecf, line 69: It doesn't look like you modify selected, colorway, or handleClick, so these can be `const`s. Or you could just directly pass them to PuckBox. 

Same comment for the `handleClick` above

PR #51, opened 2020-05-01
Add RabbitMQ

morganecf:
This works for me locally, though of course I'm not using multiple workers. This only runs rabbitmq if you're in production mode, so this shouldn't affect any local development, and you can still develop locally without installing any rabbitmq dependencies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,8 @@ npm install
 
 ### Running the app locally 
 ```bash
+# Make sure your virtual env is activated
+source venv/bin/activate
 cd app/
 
 # Export env variables. Use setup_prod.sh for prod
@@ -39,7 +41,7 @@ bash run.sh
 
 The app should now be running on localhost:5000.
 
-### Querying DB
+### Querying the DB
 Using sqlite directly
 ```
 sqlite3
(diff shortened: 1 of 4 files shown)


morganecf:
From docs:

> There are two requirements to use multiple Flask-SocketIO workers: The load balancer must be configured to forward all HTTP requests from a given client always to the same worker. This is sometimes referenced as “sticky sessions”. For nginx, use the ip_hash directive to achieve this. Gunicorn cannot be used with multiple workers because its load balancer algorithm does not support sticky sessions. Since each of the servers owns only a subset of the client connections, a message queue such as Redis or RabbitMQ is used by the servers to coordinate complex operations such as broadcasting and rooms.

> If eventlet or gevent are used, then monkey patching the Python standard library is normally required to force the message queue package to use coroutine friendly functions and classes.

```import eventlet
eventlet.monkey_patch()
```

We are currently using eventlet, but do we need to be?

hndrewaall:
@morganecf what features do we actually want rabbitmq for? I know we'll want it for Celery, but do we need broadcasting/rooms?

hndrewaall:
I think what you quoted points out what might be our actual problem, which is consistent routing. I will enable `ip_hash` and see if that mitigates

hndrewaall:
re eventlet/gevent, yes. but I honestly can't remember the details atm wrt uwsgi. can always fix later.

PR #84, opened 2020-05-03
Unfuck the Map

morganecf:
The client now:

1. Sends a request once after the user logs in to fetch the current global state of rooms / users per room
2. Uses websockets to sync the global client state with the global server state (only active users, filtering by `user.last_seen`). For example: if a user leaves a room, this will send a socket update to flask, which updates the user location table and then broadcasts a message to all clients with a message containing the full state of active users. We use full state instead of diffs as a failsafe to prevent drift between client & server. 
3. Turns the map into a modal, so that you no longer have to leave the room to see the map.
4. Turns `user.avatar` into an object `{ type, color }` instead of an array so that we can semantically access the values by name instead of index. I was running into some errors with clashing `user.avatar` types in the client - sometimes it was showing up as `normal-puck` and other times `['normal', 'puck']`, so I decided to normalize this data structure.

### DIFF
--- a/app/client/jsx/Map.jsx
+++ b/app/client/jsx/Map.jsx
@@ -4,51 +4,23 @@ import { connect } from 'react-redux'
 import reducers from './reducers.jsx'
 import MapVisualization from './MapVisualization.js'
 import MapRoomInfo from './MapRoomInfo.jsx'
-import { HttpApi } from './WebAPI.jsx'
 
 class Map extends Component {
     constructor(props) {
         super(props)
-
         this.state = {
             highlighted: null,
             users: []
         }
-
-        this.httpApi = new HttpApi()
-
-        this.props.socketApi.on('user-left-room', this.fetchUsers.bind(this))
-        this.props.socketApi.on('user-entered-room', this.fetchUsers.bind(this))
-        this.props.socketApi.on('user-disconnected', this.fetchUsers.bind(this))
     }
 
-    async fetchUsers() {
-        const { success, users } = await this.httpApi.getUsers()
-        if (success) {
-            const usersByRoom = _.reduce(users, (result, user) => {
-                (result[user.room] || (result[user.room] = [])).push(user)
-                return result
-              }, {});
-
-            this.setState({
-                users,
-                usersByRoom
-            })
-        }
-    }
-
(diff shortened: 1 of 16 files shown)


app/client/jsx/Room.jsx
```diff
@@ -202,6 +198,9 @@ class Room extends Component {
         const isMapUnlocked = _.has(this.roomTypesWithMap, room.type) && (visitedRooms >= mapUnlockThreshold)
         const showMapTooltip = isMapUnlocked && visitedRooms == mapUnlockThreshold
 
+        // Allows modal to have access to react components and state
+        Modal.setAppElement('.app')
```
morganecf, line 202: Maybe we should have this be a constant somewhere

app/client/jsx/Map.jsx
```diff
@@ -4,51 +4,23 @@ import { connect } from 'react-redux'
...
         }
-
-        this.httpApi = new HttpApi()
-
-        this.props.socketApi.on('user-left-room', this.fetchUsers.bind(this))
-        this.props.socketApi.on('user-entered-room', this.fetchUsers.bind(this))
-        this.props.socketApi.on('user-disconnected', this.fetchUsers.bind(this))
     }
 
-    async fetchUsers() {
```
dsshimel, line 25: Instead of sending an HTTP request to get all the users, why not send all the users down as part of the socket event? Then you can avoid an extra round trip to the server (and possibly some database queries as well).

app/jitsi/models.py
```diff
@@ -51,28 +58,26 @@ def enter_room(cls, user_id, room_name):
...
+                room = Room.query.filter_by(id=location.room_id).first().name
+                user_lists[room].append(user_dict)
+            else:
+                user_lists['hallway'].append(user_dict)
+        return user_lists
+
+    def to_dict(self):
+        user_dict = super().to_dict()
+        user_dict['avatar'] = self.get_avatar()
+        return user_dict
```
seangibat, line 80: Approving the PR but talked with Morgane about improving this later on my setting my relationships between the SQLAlchemy models that would allow us to do all this in one query w/ automagic SQLAlchemy joins.

PR #95, opened 2020-05-08
Add LocalStorage abstraction and use it to do basic sessions

SeanMcGrath:
now we will keep you "signed in" across refreshes, as long as you do something that makes a room render at least every 10min

### DIFF
--- a/app/client/jsx/LocalStorage.jsx
+++ b/app/client/jsx/LocalStorage.jsx
@@ -1,0 +1,37 @@
+const ttl = 10 * 60 * 1000 // 10 mins
+
+export default class LocalStorage {
+
+    static set(key, value) {
+        const now = new Date()
+        const item = {
+            value: value,
+            expiry: now.getTime() + ttl
+        }
+
+        localStorage.setItem(key, JSON.stringify(item))
+    }
+
+    static get(key) {
+        const itemStr = localStorage.getItem(key)
+        if (!itemStr) {
+            return null
+        }
+
+        const item = JSON.parse(itemStr)
+        const now = new Date()
+        if (now.getTime() > item.expiry) {
+            localStorage.removeItem(key)
+            return null
+        }
+        return item.value
+    }
+
+    // reset TTL
+    static touch(key) {
+        const currentValue = LocalStorage.get(key)
+        if (currentValue) {
+            LocalStorage.set(key, currentValue)
+        }
+    }
+}
--- a/app/client/jsx/LocalStorage.jsx
+++ b/app/client/jsx/LocalStorage.jsx
@@ -1,29 +1,44 @@
 const ttl = 10 * 60 * 1000 // 10 mins
+const topLevelKey = "APP"
+
+function getTopLevel() {
+    const topLevel = localStorage.getItem(topLevelKey)
+    return topLevel ? JSON.parse(topLevel) : null
+}
 
 export default class LocalStorage {
 
     static set(key, value) {
         const now = new Date()
         const item = {
-            value: value,
+            value,
             expiry: now.getTime() + ttl
         }
 
-        localStorage.setItem(key, JSON.stringify(item))
+        const current = getTopLevel() || {}
+        current[key] = item
+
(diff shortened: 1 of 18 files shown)


app/client/jsx/LocalStorage.jsx
```diff
@@ -0,0 +1,37 @@
+const ttl = 10 * 60 * 1000 // 10 mins
+
+export default class LocalStorage {
+
+    static set(key, value) {
+        const now = new Date()
+        const item = {
+            value: value,
```
morganecf, line 8: Could just be `value`

app/client/jsx/Navigation.jsx
```diff
@@ -4,6 +4,7 @@ import { connect } from 'react-redux'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import { faArrowUp, faArrowDown, faArrowLeft, faArrowRight, faMap } from '@fortawesome/free-solid-svg-icons'
 import { Avatars } from './avatars.jsx'
+import { LocalStorage } from './LocalStorage.jsx'
```
morganecf, line 7: Should this be imported here?

app/client/jsx/Room.jsx
```diff
@@ -39,17 +42,41 @@ class Room extends Component {
...
+            room,
+            entered: false,
+            users: []
+        }
+
+        this.httpApi = new HttpApi()
+    }
+
+    async fetchUsersForRoom(room) {
+        const { success, users } = await this.httpApi.getUsers(room || this.state.room)
```
morganecf, line 55: Ideally I'd merge my PR first, and then you'd change this to a `httpApi.getUsers()` call for all of the users. We only fetch the whole state once now and no longer make requests for users per room.

app/client/jsx/Room.jsx
```diff
@@ -179,7 +197,27 @@ class Room extends Component {
...
+        return {
+            mapVisible: isMapUnlocked,
+            showMapTooltip: showMapTooltip
+        }
+    }
+
     render() {
+        if (Object.keys(this.props.rooms).length === 0) {
+            // room config not loaded yet, just drop a placeholder
+            return <div></div>
```
morganecf, line 217: I think you can also just return (undefined) without returning a placeholder

app/client/jsx/Welcome.jsx
```diff
@@ -34,23 +46,28 @@ class Welcome extends Component {
...
-            })
-            this.setState({ redirect: '/party' })
+            }
+            this.handleUserSelected(user)
+            this.setState({redirect: '/party'})
         }
     }
 
+    handleUserSelected(user) {
+        LocalStorage.set("USER", user)
```
morganecf, line 63: So this only sets user in local storage once "party" is clicked, right?

app/client/jsx/Room.jsx
```diff
@@ -179,7 +197,27 @@ class Room extends Component {
...
     }
 
+    computeMapState(room) {
+        const mapAlreadyUnlocked = LocalStorage.get("MAP_UNLOCKED")
+        const mapUnlockThreshold = 3
+        const visitedRooms = Object.values(this.props.visited).filter(x => x).length
+        const isMapUnlocked = mapAlreadyUnlocked ||
+            (_.has(this.roomTypesWithMap, room.type) && (visitedRooms >= mapUnlockThreshold))
+        const showMapTooltip = !mapAlreadyUnlocked && isMapUnlocked
+        LocalStorage.set("MAP_UNLOCKED", isMapUnlocked)
```
morganecf, line 207: I think it might be better to prefix all of our `localStorage` keys with `JITSI_APP` or something app-specific, or wrap them all in an app-specific object. It seems like right now if I print out `localStorage` I'll see `USER` and `MAP_UNLOCKED` and whatever else we're throwing in there which feels a little messy

app/client/jsx/Welcome.jsx
SeanMcGrath: yes, this is called either when "party" is clicked or when someone with a "session" hits the welcome page

PR #97, opened 2020-05-09
Fix dev environment, misc. fixes

hndrewaall:
- Fix dev flask env
- Add separate `flask run-eventlet` command
- Change dev port to 3000
- Update docker registry URL
- Fix scripts
- Linter noise I can't be fucked to clean up

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -1,6 +1,9 @@
 #! /bin/bash
 
-source setup_prod.sh
-flask create-db
-FLASK_DEBUG=1 PYTHONUNBUFFERED=1 flask run-prod
+export FLASK_APP="manager.py"
+export FLASK_ENV=production
+export FLASK_DEBUG=1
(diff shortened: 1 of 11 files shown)


start.sh
```diff
@@ -1,6 +1,9 @@
...
 
-source setup_prod.sh
-flask create-db
-FLASK_DEBUG=1 PYTHONUNBUFFERED=1 flask run-prod
+export FLASK_APP="manager.py"
+export FLASK_ENV=production
+export FLASK_DEBUG=1
+export PYTHONUNBUFFERED=1
+export FLASK_RUN_PORT=80
+flask run-eventlet
```
morganecf, line 8: You had `flask create-db` before this, do you want to keep that here?

hndrewaall: Nope, I don't want restarting workers to wipe the DB

PR #103, opened 2020-05-09
Update DJ & Art for 05/09 Party

morganecf:


app/client/jsx/Room.jsx
```diff
@@ -71,7 +71,7 @@ class Room extends Component {
         const jitsiData = {
             displayName: this.props.user.username,
             avatar: this.props.user.avatar,
-            roomName: roomData.name,
+            roomName: this.state.room,
```
morganecf, line 74: Use the room key instead of the room name to generate jitsi urls, since names can have non-standard characters in them

Issue #100, opened 2020-05-09
More efficient sqlalchemy models


morganecf:
Per @seangibat's suggestions -- `get_active_users_by_room` should be doing something like:

```python
for user in users:
   user_dict = user.to_dict()
   if user.room:
      user_lists[room.name].append(user_dict)
   else:
      user_lists['hallway'].append(user_dict)
```

Instead of querying multiple tables

seangibat:
Closed by #109.

PR #109, opened 2020-05-10
Improve DB relationships and queries

seangibat:
Addresses #100.

- Adds a one-to-many `relationship()` property between User and Room to give us automagic joins and more efficient queries
- Removes the User-Room join table `UserLocation` - shouldn't need it in a one-to-many relationship (https://docs.sqlalchemy.org/en/13/orm/basic_relationships.html)
- Modifies `get_active_users_by_room()` to make use of changes

### DIFF
--- a/app/jitsi/events.py
+++ b/app/jitsi/events.py
@@ -27,7 +27,7 @@ def on_ping(message):
 
 @socketio.on('leave-room')
 def on_leave_room(message):
-    User.leave_room(message['user']['userId'], message['room'])
+    User.leave_room(message['user']['userId'])
     broadcast_state()
 
 
--- a/app/jitsi/events.py
+++ b/app/jitsi/events.py
@@ -27,7 +27,7 @@ def on_ping(message):
 
 @socketio.on('leave-room')
 def on_leave_room(message):
-    User.leave_room(message['user']['userId'])
+    User.leave_room(message['user']['userId'], message['room'])
     broadcast_state()
 
 
--- a/app/jitsi/models.py
+++ b/app/jitsi/models.py
@@ -50,9 +50,15 @@ class User(db.Model, SerializerMixin):
     @classmethod
     def leave_room(cls, user_id, room_name):
         user = User.query.filter_by(id=user_id).first()
+
+        if not user.room:
+            # Navigated from one hallway to another.
+            return
+
         if user.room.name != room_name:
             logger.warning(f"User {user.id} ({user.username}) tried to leave room {room_name} but was in room {room.name}")
             return
+
         user.room = None
         db.session.commit()
 
--- a/app/jitsi/models.py
+++ b/app/jitsi/models.py
@@ -1,6 +1,7 @@
 from . import db
 from datetime import datetime
 from collections import defaultdict
+from sqlalchemy import UniqueConstraint
 from sqlalchemy_serializer import SerializerMixin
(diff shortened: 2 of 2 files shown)


app/jitsi/models.py
```diff
@@ -97,18 +94,10 @@ class Room(db.Model, SerializerMixin):
...
 
-    def __repr__(self):
-        return 'Room {0}'.format(self.name)
-
-
-class UserLocation(db.Model, SerializerMixin):
-    __tablename__ = 'user_locations'
-    id = db.Column(db.Integer, primary_key=True)
-    room_id = db.Column(db.Integer, db.ForeignKey('rooms.id'))
-    user_id = db.Column(db.Integer, db.ForeignKey('users.id'), index=True)
```
seangibat, line 108: We only need a join table if users can be in more than one room.

app/jitsi/models.py
```diff
@@ -38,38 +41,32 @@ def create(cls, username, avatar):
...
-        user_location = UserLocation(user_id=user_id, room_id=room.id)
-        db.session.add(user_location)
+        user = User.query.filter_by(id=user_id).first()
+        user.room = room
 
         # Make room discovered by user
-        user_room_state = UserRoomState(user_id=user_id, room_id=room.id, discovered=True)
-        db.session.add(user_room_state)
+        user_room_state = UserRoomState.find_or_create(user_id=user_id, room_id=room.id)
+        user_room_state.discovered = True
```
seangibat, line 58: `UserRoomState.find_or_create` makes it so that we don't have duplicate UserRoomStates. This is also now enforced by a DB constraint ony `UserRoomState` that only allows one record per `user,room`.

app/jitsi/events.py
```diff
@@ -27,7 +27,7 @@ def on_ping(message):
 
 @socketio.on('leave-room')
 def on_leave_room(message):
-    User.leave_room(message['user']['userId'], message['room'])
+    User.leave_room(message['user']['userId'])
```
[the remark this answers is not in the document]
seangibat: Yeah that makes sense - if the room we're trying to leave is not the one we're actually in, log it as a warning and ignore it

seangibat:
Okay it's tested

app/jitsi/models.py
```diff
@@ -39,37 +49,40 @@ def create(cls, username, avatar):
...
     @classmethod
     def leave_room(cls, user_id, room_name):
-        room = Room.query.filter_by(name=room_name).first()
-        user_location = UserLocation.query.filter_by(user_id=user_id, room_id=room.id).first()
-        if user_location:
-            db.session.delete(user_location)
-            db.session.commit()
+        user = User.query.filter_by(id=user_id).first()
+
+        if not user.room:
```
hndrewaall, line 54: don't we return `hallway` as room in some contexts? I seem to remember that. Could that cause issues here?

seangibat: `hallway` is only used in the return value `get_active_users_by_room()` for use in the map:
```
            if user.room:
                user_lists[user.room.name].append(user_dict)
            else:
                user_lists['hallway'].append(user_dict)
```
We never set the user's state as `hallway`.

hndrewaall: Okiedoke. Just worried that having two diff ways of representing the hallway state could lead to confusion down the road. But if it's fine now it may not be worth refactoring yet.

PR #112, opened 2020-05-12
Add IP to the SQL User Table

RibhiEl-Zaru:
Requires testing, as I was only able to have localhost IP (127.0.0.1) land into DB. I'm not sure if this is because I was running it all locally. Curious to see what happens if I access website on different server. 
Set up this PR to kinda open convo re testing/comments.

### DIFF
--- a/app/jitsi/main.py
+++ b/app/jitsi/main.py
@@ -8,7 +8,7 @@ main = Blueprint('main', __name__)
 @main.route('/join', methods=['GET', 'POST'])
 def join():
     params = request.get_json()['params']
-    params['ip'] = computeIp()
+    params['ip'] = compute_ip()
     user = User.create(**params)
     return jsonify(user.to_json())
 
@@ -37,7 +37,7 @@ def serve(path):
     return send_from_directory(current_app.static_folder, 'index.html')
 
 
-def computeIp():
+def compute_ip():
     headers_list = request.headers.getlist("X-Forwarded-For")
     # using the 0th index of headers_list is dangerous for stuff explained here: http://esd.io/blog/flask-apps-heroku-real-ip-spoofing.html 
     # TODO find a way to make this ALWAYS get client IP
--- a/app/jitsi/main.py
+++ b/app/jitsi/main.py
@@ -8,6 +8,7 @@ main = Blueprint('main', __name__)
 @main.route('/join', methods=['GET', 'POST'])
 def join():
     params = request.get_json()['params']
+    params['ip'] = computeIp()
     user = User.create(**params)
     return jsonify(user.to_json())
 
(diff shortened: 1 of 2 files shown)


app/jitsi/main.py
```diff
@@ -34,3 +35,12 @@ def serve(path):
     if path and not path.startswith('client'):
         return redirect(url_for('main.serve'))
     return send_from_directory(current_app.static_folder, 'index.html')
+
+
+def computeIp():
```
seangibat, line 40: use camel_case in Python

app/jitsi/main.py
```diff
@@ -34,3 +35,12 @@ def serve(path):
...
+    headers_list = request.headers.getlist("X-Forwarded-For")
+    # using the 0th index of headers_list is dangerous for stuff explained here: http://esd.io/blog/flask-apps-heroku-real-ip-spoofing.html 
+    # TODO find a way to make this ALWAYS get client IP
+    user_ip = headers_list[0] if headers_list else request.remote_addr
```
seangibat, line 44: Is `X-Forwarded-For` only present when there's a proxy? I see in the link/StackOverflow that they're also checking `remote_addr` like so:
```
if not request.headers.getlist("X-Forwarded-For"):
   ip = request.remote_addr
else:
   ip = request.headers.getlist("X-Forwarded-For")[0]
```

app/jitsi/main.py
morganecf: Sean...this is `camelCase` !

seangibat: woops sorry! I meant snake case 😅

hndrewaall: you want `snake_case`

hndrewaall: 🐍

app/jitsi/main.py
RibhiEl-Zaru: @seangibat yup! Also, when I ran locally, request.headers['X-Forwarded-For'] was always empty, and remote_addr was used. This value was always 127.0.0.1

seangibat: ohhh I'm just blind, I didn't see you're doing a ternary if! my b :smile_cat:

app/jitsi/main.py
RibhiEl-Zaru: addressed

app/jitsi/main.py
```diff
@@ -34,3 +35,12 @@ def serve(path):
     if path and not path.startswith('client'):
         return redirect(url_for('main.serve'))
     return send_from_directory(current_app.static_folder, 'index.html')
+
+
+def compute_ip():
+    headers_list = request.headers.getlist("X-Forwarded-For")
```
hndrewaall, line 41: in this link it mentions a fix committed to `werkzeug`: https://github.com/pallets/werkzeug/commit/cdf680222af293a2c118d8d52eecfd7b0c566e14

It's not clear to me that this helper is still available when flask is run using things other than `werkzeug` for WSGI (like `eventlet`, as we do in prod). But even if it isn't, the basic idea is trivial to implement: pass in `num_proxies` as an argument, and use that to index into the header value. Let's do this here? If we pull that from the Config object and define it as 0 in the dev config and 1 in the prod config, things should work.

hndrewaall:
Closing in favor of #170

PR #114, opened 2020-05-15
First steps towards more configurability

SeanMcGrath:
the basic idea is that there's a static `Config` object accessible from which you can pull values which are set in a top-level `config.json`

in production (read as "not localhost") overrides in `production.json` take effect.


app/config.json
```diff
@@ -0,0 +1,484 @@
+{
+	"baseUrl": "http://localhost:5000",
+	"debug": true,
+	"mapUnlockThreshold": 3,
+	"useLocalSessions": true,
+ 	"rooms": {
```
morganecf, line 6: Idk, would it be useful to have rooms in a separate json? This config file might get pretty big

app/jitsi/main.py
```diff
@@ -14,7 +14,8 @@ def join():
 @main.route('/rooms')
 def get_rooms():
     basedir = current_app.config.get('BASE_DIR')
-    rooms = json.load(open(os.path.join(basedir, 'rooms.json')))
+    config = json.load(open(os.path.join(basedir, 'config.json')))
```
morganecf, line 17: I wonder if we could specify the paths to dev/prod configs in `config.py`?

app/config.json
```diff
@@ -0,0 +1,484 @@
...
+			},
+			"doorTint": "black"
+		},
+		"bye": {
+			"name": "Bye",
+			"type": "redirect",
+			"route": "/bye"
+		}
+	}
+}
```
morganecf, line 484: github complains when there's no new line at EOF

app/client/jsx/Config.jsx
```diff
@@ -0,0 +1,9 @@
+import baseConfig from '../../config.json'
+import prodConfig from '../../production.json'
+
+// values in prodConfig override values in baseConfig when not running locally
+const hostname = window && window.location && window.location.hostname
+const development = hostname && hostname.includes('localhost')
+const config = !development ? Object.assign(baseConfig, prodConfig) : baseConfig
```
morganecf, line 7: Yay! I was flustered by keeping this in `WebAPI.jsx` - thank you

app/client/jsx/Room.jsx
```diff
@@ -166,9 +169,11 @@ class Room extends Component {
 
     computeMapState(room) {
         const mapAlreadyUnlocked = LocalStorage.get("MAP_UNLOCKED")
-        const mapUnlockThreshold = 3
+        const mapUnlockThreshold = parseInt(Config.mapUnlockThreshold) || 3
```
morganecf, line 172: This `3` makes me think we should have a _generic_ config that sets defaults and is overridden by a user-specified config. Default values could be specified in another config file or programmatically, wherever we parse the user config

app/config.py
```diff
@@ -1,7 +1,20 @@
...
 
+def get_json_dict(path):
+    try:
+        with open(path) as file:
+            return json.load(file)
+    except:
+        return dict()
+
+def make_merged_cfg(paths):
+    print(paths)
```
morganecf, line 16: Thanks for these changes! I'm excited. I think you left a `print` statement here unintentionally

PR #115, opened 2020-05-19
PuckSelect housekeeping

danielmdavis:
quality changes excluding those involving on webpack, assets, genericization

### DIFF
--- a/app/client/jsx/PuckSelect.jsx
+++ b/app/client/jsx/PuckSelect.jsx
@@ -1,37 +1,38 @@
-import React, { Component } from 'react';
-import PuckBox from './PuckBox.jsx';
+import React, { Component } from 'react'
+import PuckBox from './PuckBox.jsx'
 import { Avatars } from './avatars.jsx'
 
-
 class PuckSelect extends Component {
-    constructor(props) {
-        super(props)
-        this.state = {
-            columnOpen: false,
-            rowOpen: null,
-            avatarDesign: '',
-            avatarColorway: '',
-            opacity: this.props.opacity
-        }
-        this.onButtonClick = this.onButtonClick.bind(this)
-        this.onButtonClickReset = this.onButtonClickReset.bind(this)
-        this.handleClickDesign = this.handleClickDesign.bind(this)
-        this.handleClickColor = this.handleClickColor.bind(this)
-      }
+  constructor (props) {
+    super(props)
+    this.state = {
+      columnOpen: false,
+      rowOpen: null,
+      avatarDesign: '',
+      avatarColorway: '',
+      opacity: this.props.opacity
+    }
+    this.onButtonClick = this.onButtonClick.bind(this)
+    this.onButtonClickReset = this.onButtonClickReset.bind(this)
+    this.handleClickDesign = this.handleClickDesign.bind(this)
+    this.handleClickColor = this.handleClickColor.bind(this)
+  }
 
-  onButtonClick() {
-    this.setState({ columnOpen: true }) }
+  onButtonClick () { this.setState({ columnOpen: true }) }
 
-  onButtonClickReset() {
+  onButtonClickReset () {
     this.setState({ columnOpen: false })
     this.setState({ rowOpen: null })
     this.setState({ avatarDesign: '' })
     this.setState({ avatarColorway: '' })
     this.setState({ columnOpen: true })
   }
-  handleClickDesign(key, index) {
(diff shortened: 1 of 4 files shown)


app/client/jsx/PuckSelect.jsx
```diff
@@ -1,108 +1,98 @@
...
     this.setState({ avatarColorway: variantKey })
     this.props.handleSelect({
       type: this.state.avatarDesign,
       color: variantKey
     })
   }
 
-  render() {
+  render () {
+    console.log = console.warn = console.error = () => {}
```
SeanMcGrath, line 38: this looks mighty suspicious. won't this leak out and affect the function globally?

danielmdavis: I'd agree that it's suspicious, even distasteful. Suffice it to say it only entered the picture after many involved attempts to determine the cause, and at Morgane's suggestion.

This said, if it's cutting across scope we've got a problem

SeanMcGrath: personally i'd much rather have some garbage in the console than to clobber this function globally

seangibat: what is it that this is trying to solve? I forget

danielmdavis: one of these. of unprecedented tenacity 
Screen Shot 2020-05-19 at 11 55 12 AM

seangibat: I think you need to set a unique key on all the elements inside PuckBox, maybe
https://stackoverflow.com/questions/28329382/understanding-unique-keys-for-array-children-in-react-js

morganecf: I suggested silencing it. You can silence it with a try / catch

app/client/jsx/Welcome.jsx
```diff
@@ -8,127 +8,109 @@ import { HttpApi } from './WebAPI.jsx'
...
+  }
 
-        let name_opacity = 'form-fade'
-        let avatar_opacity = 'form-fade'
-        let party_opacity = 'form-fade-party'
-        if (this.state.username === null) { name_opacity = 'form' }
-        if (this.state.username) { avatar_opacity = 'form' }
-        if (this.state.username && this.state.avatar) { party_opacity = 'form-party' }
+  render () {
+    if (this.state.redirect) { return <Redirect to={this.state.redirect} /> }
```
morganecf, line 91: Can this be broken up into two lines please?

app/client/jsx/Welcome.jsx
```diff
@@ -8,127 +8,109 @@ import { HttpApi } from './WebAPI.jsx'
...
+    const nameOpacity = (this.state.username === null) ? 'form' : 'form-fade'
+    const avatarOpacity = (this.state.username) ? 'form' : 'form-fade'
+    const partyOpacity = (this.state.username && this.state.avatar) ? 'form-party' : 'form-fade-party'
```
morganecf, line 94: Is there a way to combine these? It looks like you evaluate `this.state.username` several times; maybe this can be done once

app/client/styles/components/_Welcome.scss
```diff
@@ -27,6 +27,18 @@ input[type="button"] {
...
+
+.text-entry {
+	padding: 6px 10px !important;
+  border: 2px solid #D6D3CD !important;
+  background-color: #D6D3CD !important;
+  border-radius: 7px !important;
+  color: #1B1E1F !important;
+  outline: none !important;
+  box-shadow: none !important;
+  font-size: 20px !important;
```
morganecf, line 39: Can the indenting here be consistent?

app/client/jsx/PuckSelect.jsx
morganecf: @seangibat it seems like a more intractable problem than that for whatever reason

seangibat: It's tricky to make sure each div has a unique key, but if you do that, you won't get the warning. I tried it here and it worked: https://github.com/morganecf/jitsi-party/pull/116 (But this is really messy demonstration and it might be that not all of it is necessary.)

seangibat: On further testing I realized the only thing you need to do is set the key on the main outer div! I don't quite understand why.

morganecf: I'll defer to @danielmdavis on this one! That fix looks good to me.

PR #116, opened 2020-05-19
Fix unique key warning

seangibat:


danielmdavis:
I'd tried inner divs, outer divs, creative naming at different levels in case the problem was key name collision! I wonder what you did differently

PR #122, opened 2020-05-29
Modernizing JS

feitelogram:
Testing linting our React app, adding ES6 props destructuring and getting rid of "export default" anonymous functions as they are not semantically preferred.


hndrewaall:
Apologies if obvious, but how does one invoke the linter in the future? Can we add to docs?

app/client/jsx/Exit.jsx
```diff
@@ -1,32 +1,26 @@
...
-    }
+  render() {
+    // Support either a single "goodbye" avatar or a map of color -> image
+    const goodbyeImgUrl =
+            (Config.specialAvatars && Config.specialAvatars.goodbye) ?
+                (Config.specialAvatars.goodbye[this.props.user.avatar.color] || Config.specialAvatars.goodbye) :
+                null;
 
-    render() {
-        localStorage.clear() // log user out
```
SeanMcGrath, line 11: we seem to have dropped this

feitelogram: adding

danielmdavis:
So I see that, for instance on door.jsx, it looks like you've gone through and added a lot of semicolons. it's my understanding that most major style guides have moved away from them (StandardJS, AirBnB, etc.)

feitelogram:
Yeah I’d want to go through it with you. Do you have any time this weekend?

hndrewaall:
closed automatically due to history rewrite. can reopen when pr branch is rebased

PR #124, opened 2020-05-31
Remove baseUrl from config

morganecf:
I don't think we need localhost:port or the production url hardcoded in the config since we can just introspect it from the window.


hndrewaall:
More importantly, having it actually sucks and breaks a lot of things

PR #126, opened 2020-05-31
Genericize map stats

morganecf:
Config now controls the stats displayed on the map. The three default stats (# in house, # in rooms, # roaming) can be switched on or off and the text/emoji can be customized. More stats can be added in the `additionalStats` list, which counts up users in rooms matching a given list of rooms or a regular expression. 

Let's say the following additional stats are added:
```json
"additionalStats": [
	{
		"text": "{0} doing activity",
		"match": ".*Room",
		"emoji": "🎯"
	},
	{
		"text": "{0} in closet",
		"match": ["closet"],
		"emoji": "🗝"
	}
]
```

The first stat will match on any room key ending in `Room`, and the second will match on `closet`, resulting in the following header, where the user is currently in the Game Room:

Screen Shot 2020-05-31 at 1 57 28 PM

### DIFF
--- a/app/client/jsx/Map.jsx
+++ b/app/client/jsx/Map.jsx
@@ -4,6 +4,7 @@ import { connect } from 'react-redux'
 import reducers from './reducers.jsx'
 import MapVisualization from './MapVisualization.js'
 import MapRoomInfo from './MapRoomInfo.jsx'
+import Config from './Config.jsx'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import { faTimes } from '@fortawesome/free-solid-svg-icons'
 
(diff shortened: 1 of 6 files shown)


app/config/base.json
```diff
@@ -18,5 +18,25 @@
...
+			},
+			"numberInRooms": {
+				"display": true,
+				"text": "{0} in rooms",
+				"emoji": "🚪"
+			},
+			"numberInHallways": {
+				"display": true,
+				"text": "{0} roaming",
+				"emoji": "🌓"
```
morganecf, line 37: We could maybe just combine the text and emoji keys:

```json
{
   "emoji": "🚪  {0} in rooms"
}
```

app/config/overrides/cabin.config.json
```diff
@@ -49,5 +49,14 @@
 	    "red": "./js/images/puck/thumb-red.png",
 	    "orange": "./js/images/puck/thumb-orange.png"
 	  }
+	},
+	"map": {
+		"additionalStats": [
+			{
+				"text": "{0} taking bathroom break",
```
morganecf, line 56: Maybe these should also have a `display` key for convenience, in case a config manager wants to toggle things on and off without having to remove / readd stats

app/client/jsx/Map.jsx
```diff
@@ -55,35 +56,52 @@ class Map extends Component {
...
+            emoji: config[key].emoji
+        }))
+
+        config.additionalStats.forEach(({ text, match, emoji }) => {
+            const rooms = _.isArray(match) ?
+                _.intersection(Object.keys(users), match) :
+                Object.keys(users).filter(room => new RegExp(match).test(room))
+            const count = _.sumBy(rooms, room => users[room].length)
+            mapStats.push({
+                text: text.replace('{0}', count),
```
SeanMcGrath, line 92: FYI if we're not doing multiple format parameters, `{0}` might be a little misleading

app/config/base.json
SeanMcGrath: i am in favor of this

app/client/jsx/Map.jsx
morganecf: Agreed...I tried `{}` but it looked a little weird to me. There must be a standard for this. Do you have a suggestion? Maybe `%` or `%s`? Wildcard?

seangibat: what about `{count}`?
`"{count} taking bathroom break"`
`text.format(count=count)`

Issue #128, opened 2020-06-01
Improve adventures architecture


morganecf:
In `main.py` we flatten the adventures into rooms, treating each node in the adventure as a room. This makes it difficult to add high-level features to each adventure (such as an audio track) -- we'd have to specify it in every single adventure node and update `main.get_rooms`.

PR #129, opened 2020-06-01
Add some basic audio

morganecf:
Adds the ability to play an audio track when you enter a room. Looks like this:

Screen Shot 2020-06-01 at 1 47 25 PM

Users can turn the audio on or off. This state is persisted for each room, so if the user turns off audio for one room, it will not auto play when they go back to that room. Whether or not audio autoplays is also configurable.

I currently don't mute the audio when you actually ENTER a room, which is probably bad UX if we intend to place audio in social rooms, _but_ I mostly envision using this for adventures right now and maybe the art gallery / solo closets, so idk if this is a problem. Andrew and I had also discussed being able to trigger audio events with celery, so this is a starting point for that.

The audio I'm currently using is just for demonstration purposes. I'll change it up before the next party.


app/client/styles/components/_button.scss
```diff
@@ -18,14 +18,18 @@ input[type="button"] {
 	}
 
 	&:hover,
-	&:focus {
+	&.focused {
```
morganecf, line 21: @SeanMcGrath I removed the focus rules because I don't think they're useful right now and they mostly result in a button already looking selected when we switch between rooms.

app/client/jsx/AudioPlayer.jsx
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    render() {
+        const buttonClasses = [
+            'audio-button'
+        ]
+        if (!this.state.paused) {
+            buttonClasses.push('focused')
+        }
+        if (this.props.hide) {
```
SeanMcGrath, line 69: if only JS has `Option`

app/jitsi/main.py
```diff
@@ -13,19 +13,21 @@ def join():
 
 @main.route('/rooms')
 def get_rooms():
-    basedir = current_app.config.get('BASE_DIR')
     rooms = current_app.config['ROOMS']
     adventures = current_app.config['ADVENTURES']
     for adventure in adventures.values():
+        config = adventure.get('config', {})
```
SeanMcGrath, line 19: this way of setting up rooms/adventures still feels pretty brittle to me, but i don't claim to have a better solution on hand...

morganecf: I agree -- this `config` stuff is, in my mind, temporary. I am ok with the way rooms are set up but I definitely want to reform the adventures. I made a ticket for this yesterday: https://github.com/morganecf/jitsi-party/issues/128 

Not entirely sure what the solution is but I at least want to think about it some more

PR #132, opened 2020-06-06
Event View

morganecf:
Adds ability to display custom party events. Events are added to an `overrides/events.json` config.

Calendar button shows up next to map if there are events:
Screen Shot 2020-06-06 at 3 04 27 PM

Clicking on the button opens up an overlay:
Jun-06-2020 15-03-46


app/client/jsx/Room.jsx
```diff
@@ -221,14 +230,22 @@ class Room extends Component {
...
                 <Beforeunload onBeforeunload={this.handleBeforeUnload.bind(this)} />
                 <Modal
                     isOpen={this.state.showMap}
                     onAfterOpen={this.handleOpenMap.bind(this)}
                     onRequestClose={this.handleCloseMap.bind(this)}>
                         <Map onRoomClick={this.onSwitchRoom.bind(this)} handleCloseMap={this.handleCloseMap.bind(this)}></Map>
                 </Modal>
+                <Modal
+                    isOpen={this.state.showEvents}
+                    onAfterOpen={this.handleOpenEvents.bind(this)}
```
SeanMcGrath, line 244: am i reading this correctly that clicking the button triggers `handleOpenEvents` but we're also calling in `onAfterOpen`?

app/client/styles/components/_eventList.scss
```diff
@@ -0,0 +1,59 @@
...
+.ReactModal__Content.event-modal {
+    position: absolute;
+    width: 50%;
+    max-height: 85%;
+    left: 25%;
+    top: 40px;
+    border-radius: 5px;
+    outline: none;
+    overflow-y: auto;
+}
```
SeanMcGrath, line 59: newline at EOF

app/config.py
```diff
@@ -44,6 +49,13 @@ def ADVENTURES(self):
             return self.merged_cfg['adventures']
         except:
             return dict()
+
+    @property
+    def EVENTS(self):
+        try:
+            return self.merged_cfg['events']
```
SeanMcGrath, line 56: i think it's my bad for doing `try/except` here originally, can we not do `merged_cfg.get('events', dict())`?

morganecf: yes :)

app/client/jsx/Room.jsx
morganecf: Hm...yeah that doesn't make sense. I don't think we need that callback.

morganecf: FYI I removed this callback but something funky happens when I do a similar thing for the map modal. Will look into this later.

PR #135, opened 2020-06-07
Themes - first take

SeanMcGrath:
This PR sets the app up with an scss color palette (which copies the google material design color palette) and defines all colors in the application with reference to these canonical colors, mapped via theme scss file.

In theory one can now meaningfully restyle the app via palette swap. This PR does not enable theme "hotswapping", since the colors are defined globally at buildtime


app/client/styles/components/_Navigation.scss
```diff
@@ -6,7 +6,7 @@
     display: flex;
     align-items: center;
     padding: 8px 50px;
-    background-color: #191b1c;
+    background-color: $background-color-dark;
```
morganecf, line 9: Oh, good to know that these variables are all global. I didn't realize that before.

app/config/base.json
```diff
@@ -4,7 +4,7 @@
 	"useLocalSessions": true,
 	"welcomePage": {
 		"avatarSelectionEnabled": true,
-		"backgroundColor": "black",
+		"backgroundColor": "#121212",
```
morganecf, line 7: So in a next step, would this backgroundColor plus additional styling be specified in an scss file that overrides `_default.scss`?

SeanMcGrath: yeah, something along those lines

app/client/styles/abstracts/_palette.scss
```diff
@@ -0,0 +1,68 @@
...
+
+$palette: map-merge($custom-palette, $material-palette);
+
+@function color($key) {
+  @if map-has-key($palette, $key) {
+    @return map-get($palette, $key);
+  }
+
+  @warn "Unknown `#{$key}` in $palette.";
+  @return null;
```
morganecf, line 66: Does this mean we can't use themes that aren't based on the material design color palette at this moment?

SeanMcGrath: nah you can totally define custom colors. I just think it's useful to have a canonical palette

SeanMcGrath: the theme doesn't require values to be specified in terms of this function

app/client/styles/abstracts/_palette.scss
```diff
@@ -0,0 +1,68 @@
...
+         #{$color}-#{$lighten}: lighten($value, $percent),
+         #{$color}-#{$darken}:  darken($value, $percent)
+      ));
+    }
+
+    // middle eg 500
+    $colors: map-merge($colors, (#{$color}-50: lighten($value, 38.7)));
+    $colors: map-merge($colors, (#{$color}-#{$variations/2*100}: $value));
+  }
+  @return $colors;
```
morganecf, line 50: Does this mean we can only enter values such as `purple-500` rather than just `purple`?

app/config/base.json
morganecf: Should we remove this from the default json config since it's now specified in the default scss?

app/client/styles/abstracts/_palette.scss
SeanMcGrath: yes you have to specify the "weight"

app/client/styles/main.scss
```diff
@@ -4,27 +4,33 @@
...
 
-// 4. Layout-related sections
+// 5. Layout-related sections
 @import
   'layout/header',
   'layout/footer';
 
-// 5. Components
+// 6. Components
 @import
```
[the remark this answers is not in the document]
SeanMcGrath: yes, you could drop in more theme styling here

PR #139, opened 2020-06-12
Notifications & Actions

morganecf:
SORRY THIS IS KINDA LARGE!! 

### Summary

There are 2 types of notifications:
1. Regular toast notifications (text-based)
2. Audio announcements

The actions let us **schedule** things like notifications, announcements, custom modals, room switches, theme switches, shuffling users across rooms --- so we can do an early rudimentary version of the chat roulette we've been talking about.

All of this logic is purely client-side! In the future, if we want to trigger random global actions or notifications from an admin panel, we'll use something like celery + this code.

### Example
The gif below results from the following action config:
```json
{
            "type": "NOTIFICATION",
            "secondsAfterJoin": 2,
            "notification": {
                "text": "Hello there! Welcome to the party.",
                "image": "js/images/puck/book-orange.png"
            }
        },
        {
            "type": "CHANGE_THEME",
            "time": "2020-06-10T23:59:00-04:00",
            "triggerAfterTime": true,
            "className": "midnight"
        },
        {
            "type": "SHUFFLE_ROOM",
            "secondsAfterJoin": 6
        }
```

Jun-11-2020 21-54-14
1. The theme is changed instantaneously, since we joined the app after the trigger time and `triggerAfterTime` is set to true.
2. A notification pops up 2 seconds after load
3. A custom action is scheduled for 6 seconds in. It triggers an alert and sends the user to a random room

For the most part I think we'll use timestamps for our purposes, not `secondsAfterJoin`, but it's useful to have for testing.

### Potential use cases for Saturday
Assuming this gets merged on time.
1. Notifications for events/performances
2. Portal opens up for everyone within a 5-10m timespan at midnight (`MidnightPortal`). If the user goes through it, the app theme will change and they will be taken to the starting point of an underground location that henceforth was not a part of the map.
3. If user joins after midnight, theme changes automatically

### Note on actions

A notification can be a type of action. Actions are events that happen at a given time or a given number of seconds after the user has joined the app. They are specified in the `actions.json` config. An action has a `type` that is associated with a function in `actions.js`. These are very similar to redux actions. I think we should ultimately implement these as real redux actions that can be `dispatched`, but right now **changing the redux state doesn't actually trigger changes in the Room view**, which means we can't schedule an action via redux to change the room state. For example, if we wanted to schedule an action to make everyone switch to the kitchen at midnight, it'd be nice to just do something like:
```javascript
dispatch({ type: UPDATE_CURRENT_ROOM, room: 'kitchen' })
```
However, the redux `updateCurrentRoom` action only updates the global state with the room string, and doesn't do the socket updates or background clearing and all the other stuff handled by `Room.onSwitchRoom` and `Room.onEnterRoom`. 

The way we get around this now is by passing the Room component to the action scheduler so that we can make use of its state and by extension, its props (which include the redux state). This is less than ideal but we need to address https://github.com/morganecf/jitsi-party/issues/140 before doing this the "right" way.

### Note on notifications

I thought there were benefits to isolating the logic into a context:
1. Notifications ought to be global, so if we end up adding more pages these will continue to behave globally
2. The notification context manages its own state, so we don't need to add a bunch more logic to Room
3. Child components far down the tree can use the notification context without needing to cascade down props, similar to how you can connect a particular component to the redux store. 

Ideally contexts are used with hooks, but since we're still using class components in many places, we don't currently use the custom hook I supplied in `Notifications.jsx`. When/if we transition to using only functional components, we'll be able to use that hook like so:
```javascript
import { useNotifications } from './Notifications.jsx'
...
const { add } = useNotifications()
// In some callback, trigger a notification
add({ text: 'i am notifying!' })
```

Right now we have to set the static attribute `contextType` in a class component to allow it to use `this.context`.


app/client/jsx/CustomComponents/MidnightPortal.jsx
```diff
@@ -0,0 +1,60 @@
...
+            </div>
+        )
+    }
+
+    return (
+        <div className="midnight-portal" style={style}>
+            {getInnerDiv()}
+        </div>
+    )
+}
```
morganecf, line 60: This isn't finished yet, but will be used to create the aforementioned portal and is used to demonstrate the OPEN_MODAL action

hndrewaall:
So tbc this is largely UX work with the schedule setup largely as a POC right? cuz this will always be fucked by clock skew, so ultimately we will def want even scheduled events to be triggered by a single async worker

hndrewaall:
automatically closed due to history rewrite. can be reopened when rebased but has conflicts im not qualified to address

PR #142, opened 2020-06-13
Modify specification of Jitsi server URL and noSSL option for configurability and to work w/ docker

seangibat:
- move Jitsi server URL to config
- create option to override Jitsi server URL with window.location.host/jitsi
  - use this option on dev and prod
- specify noSSL in config
  - true on dev, false on prod

### DIFF
--- a/app/client/jsx/JitsiVideo.jsx
+++ b/app/client/jsx/JitsiVideo.jsx
@@ -84,9 +84,9 @@ class JitsiVideo extends Component {
                 toolbarButtons = _.without(this.toolbarButtons, 'microphone')
             }
 
-            let domain = `${window.location.host}/jitsi/`
-            if (Config.jitsiServerUrl) {
-                domain = Config.jitsiServerUrl
+            let domain = Config.jitsiServerUrl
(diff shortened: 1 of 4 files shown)


app/client/jsx/JitsiVideo.jsx
```diff
@@ -84,9 +84,13 @@ class JitsiVideo extends Component {
...
 
-            const domain = 'party.gbre.org/jitsi/'
-            // const domain = 'meet.jit.si';
+            let domain = `${window.location.host}/jitsi/`
+            if (Config.jitsiServerUrl) {
+                domain = Config.jitsiServerUrl
+            }
+
             const options = {
+                noSSL: true,
```
hndrewaall, line 93: ideally we pull this out to the config, but this is def an improvement, so can leave for followup work

seangibat: that's easy enough--it should be `true` for local and `false` on prod?

hndrewaall: correct!

PR #146, opened 2020-07-20
Custom Auth, TST Auth, Demon Chat integration

morganecf:
I updated the jitsi config to include an optional iframe link for custom authentication. If such an option is provided, the welcome page will first ask you to authenticate before bringing you to the original welcome page. It does this by showing the iframe specified in the config. Custom iframes are stored in `js/iframes`.

In TST's use case, they use inplayer to authenticate. I wrote a small script in the html TST provided us to send an authentication event to our app once a TST-specific session token is detected. All TST-specific code exists in the iframed document (I'm thinking of this sort of as a plugin). Our app just listens for an `onCustomAuthentication` event so that it can display the original sign-in page. 

The reason we want to authenticate is so that TST users can use the demon.chat service in certain rooms. The chat is just an iframe whose src points to another `js/iframes` "plugin," `TST_chat.html`, which was provided to us by TST. 

Changelog:
* `js/iframes/TST_chat.html`: Provided by TST. This renders the demon.chat, which can be placed in any iframe room
* `js/iframes/TST_login.html`: Provided by TST. This renders the auth portal. I added a script that sends an `onCustomAuthentication` event to the parent document
* `CustomAuthWrapper.jsx`: displays whatever auth iframe is specified in the config
* `Login.jsx`: A refactor of most of what was in `Welcome.jsx`; this renders our username/avatar selection
* `Welcome.jsx`: If the config contains an auth iframe link, show that first. If the user is authenticated, show the username/avatar selection
* `config.json`: demonstrates the custom auth option
* `rooms.json`: demonstrates iframing demon.chat 

I honestly doubt we'll ever want to do any custom auth like this in the future, but this seemed like the cleanest way to keep TST logic out of the main body of code while fulfilling their requirements.


morganecf:
Test comment

morganecf:
Test

app/config/overrides/config.json
```diff
@@ -1,14 +1,18 @@
...
 		"avatarSelectionEnabled": true,
 		"backgroundImagePath": "./js/images/bauhaus_darker.jpg",
 		"enterSpaceButtonText": "Party!",
 		"avatarSelectionText": "Pick your puck",
-		"headerHtml": "<br/><div class='bauhaus'><p>This is indeed a disturbing universe.</p><h1>Cabin Weekend is Dead. Long Live Cabin Fever.</h1>"
+		"headerHtml": "<br/><div class='bauhaus'><p>This is indeed a disturbing universe.</p><h1>Cabin Weekend is Dead. Long Live Cabin Fever.</h1>",
+		"auth": {
+			"iframe": "./js/iframes/TST_login.html",
+			"descriptionHtml": "You must log in to the TST Portal in order to use demon.chat inside the house."
+		}
```
SeanMcGrath, line 15: FYI this file is mixing tabs and spaces

app/config/overrides/config.json
```diff
@@ -1,14 +1,18 @@
...
 	"startRoom": "vestibule",
 	"welcomePage": {
 		"avatarSelectionEnabled": true,
 		"backgroundImagePath": "./js/images/bauhaus_darker.jpg",
 		"enterSpaceButtonText": "Party!",
 		"avatarSelectionText": "Pick your puck",
-		"headerHtml": "<br/><div class='bauhaus'><p>This is indeed a disturbing universe.</p><h1>Cabin Weekend is Dead. Long Live Cabin Fever.</h1>"
+		"headerHtml": "<br/><div class='bauhaus'><p>This is indeed a disturbing universe.</p><h1>Cabin Weekend is Dead. Long Live Cabin Fever.</h1>",
+		"auth": {
+			"iframe": "./js/iframes/TST_login.html",
```
SeanMcGrath, line 13: it's worth discussing i guess whether this is something we want to merge into master here or whether this should go into a fork

morganecf: I'm going to keep this here as a working example, but I have another branch for TST-specific customization that I'm guessing I should probably turn into a fork?

PR #148, opened 2020-07-21
Route to test DB write

morganecf:


hndrewaall:
this can be opened when rebased but there are conflicts im not qualified to address. might not be needed anymore anyway

PR #150, opened 2020-07-21
Email a moderator

morganecf:
TST requested that users be able to send a message to a moderator. This adds an email button in the bottom lefthand corner that users can click to get to an email client:
Screen Shot 2020-07-21 at 5 46 42 PM

Sends an email to a list of moderator emails (provided in the config) that looks like:
Screen Shot 2020-07-21 at 5 47 01 PM

Please ignore the styling for now, it looks weird because the bauhaus theme is still active. I'll clean up the theming and styling in a separate PR.

### DIFF
--- a/app/client/jsx/EmailForm.jsx
+++ b/app/client/jsx/EmailForm.jsx
@@ -1,17 +1,39 @@
 import React, { useState } from 'react'
 import { HttpApi } from './WebAPI.jsx'
+import Config from './Config.jsx'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import {
+    faCircleNotch,
+    faEnvelope,
+    faExclamationCircle
+} from '@fortawesome/free-solid-svg-icons'
 
 
 export default ({ user }) => {
     const MIN_WORDS = 10
     const MAX_WORDS = 500
 
+    const STATUS = {
+        sending: {
+            label: 'Your message is being sent...',
+            icon: faCircleNotch
+        },
+        succeeded: {
+            label: 'Message sent!',
+            icon: faEnvelope
+        },
+        failed: {
+            label: `Something went wrong sending your message. Please close this box and try again, or email the moderator directly at ${Config.moderation.moderatorEmails[0]}.`,
+            icon: faExclamationCircle
+        }
+    }
+
     const httpApi = new HttpApi()
 
     const [email, setEmail] = useState('')
     const [message, setMessage] = useState('')
-    const [sending, setSending] = useState(false)
     const [sent, setSent] = useState(false)
+    const [status, setStatus] = useState('sending')
 
     const numWords = message.split(/\s+/).length - 1
     const wordCountClass = numWords < MIN_WORDS || numWords === MAX_WORDS ? 'bad-word-count' : ''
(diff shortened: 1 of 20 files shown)


hndrewaall:
Useless comment: I fucking loathe the pale pink text on white, but I assume that's CSS fuckery and actually totally irrelevant to this diff

hndrewaall:
Didn't we decide email shouldn't be optional?

app/client/jsx/EmailForm.jsx
```diff
@@ -0,0 +1,91 @@
...
+
+    const httpApi = new HttpApi()
+
+    const [email, setEmail] = useState('')
+    const [message, setMessage] = useState('')
+    const [sent, setSent] = useState(false)
+    const [status, setStatus] = useState('sending')
+
+    const numWords = message.split(/\s+/).length - 1
+    const wordCountClass = numWords < MIN_WORDS || numWords === MAX_WORDS ? 'bad-word-count' : ''
```
SeanMcGrath, line 39: it looks like on the form below you're using `numWords > MAX_WORDS` idk if that's a bug or if it works out somehow

app/client/jsx/EmailForm.jsx
```diff
@@ -0,0 +1,91 @@
...
+
+    const emailRegex = new RegExp(/^\w+([\.-]?\w+)*@\w+([\.-]?\w+)*(\.\w{2,3})+$/)
```
SeanMcGrath, line 41: how can we as professionals not use the [true email regex](http://www.ex-parrot.com/~pdw/Mail-RFC822-Address.html)

app/client/jsx/ModalFactory.jsx
```diff
@@ -0,0 +1,29 @@
...
+    const events = (
+        <EventList
+            rooms={room.props.rooms}
+            events={room.props.events}>
+        </EventList>
+    )
+    const email = (
+        <EmailForm user={room.props.user}></EmailForm>
+    )
+    return { map, events, email }[modal] || null
```
SeanMcGrath, line 28: probably don't want to create components that get thrown away, in case we end up writing one that has side effects

app/client/styles/components/_email.scss
```diff
@@ -0,0 +1,47 @@
...
+    .email-status-icon {
+        align-self: center;
+
+        svg {
+            font-size: xx-large;
+        }
+    }
+    .original-message {
+        font-size: smaller;
+        border-left: 10px solid lightgrey;
```
SeanMcGrath, line 44: ideally use a color from the color map

app/client/jsx/EmailForm.jsx
morganecf: I think it makes sense as is because we still want them to be able to send if they're AT the max word limit (but highlight the # of words in red if they're at the limit), and disable sending if they're OVER the limit.

morganecf: I noticed an issue where you can copy/paste something longer than 500w, so I also now make sure to truncate the message.

app/client/styles/components/_email.scss
morganecf: Done

app/client/jsx/ModalFactory.jsx
morganecf: What do you mean?

morganecf: I'm going to merge this so I can continue working on TST customization, but PM me!

SeanMcGrath: this piece of code appears to instantiate all three of the map, events, and email components, then discard them if none of them is the component that is asked for.

If any of these components makes network calls, manipulates some kind of state, etc, we might end up with unexpected behavior. At the very least this is kind of wasteful

app/setup.sh
```diff
@@ -1,3 +1,5 @@
 export FLASK_APP="manager.py"
 export FLASK_ENV=development
-export FLASK_RUN_PORT=3000
\ No newline at end of file
+export FLASK_RUN_PORT=3000
+
+source email_config.sh
```
bMacSwigg, line 5: This makes email_config.sh required, but it's also stripped out by the .gitignore. Should we provide an empty email_config by default, or add logic here to check for existence first?

PR #151, opened 2020-07-22
Initial Library room

bMacSwigg:
Creates a new room type and adds an example in rooms.json

Config specifies background image and imagemap areas on top. Each area include a label and list of PDFs to load. Clicking on an area loads up a modal listing the files; selecting one displays it embedded.

This PR also includes a couple other incidental changes:

- Add *.db files to the .gitignore, because it was getting really annoying that Windows kept generating thumbnail files for all the image directories
- Fix some indentation problems in rooms.json

(Apologies for the hot mess of commits. This PR is based on my master, which had been where I was naively doing all of my local setup for a party before, and I had to add commits to revert all of those commits before I could start...)


app/client/jsx/Room.jsx
```diff
@@ -83,7 +85,8 @@ class Room extends Component {
             art: <ArtRoom jitsiData={jitsiData} art={roomData.art}></ArtRoom>,
             jitsi: <JitsiVideo jitsiData={jitsiData}></JitsiVideo>,
             iframe: <IFrameRoom jitsiData={jitsiData} iframeOptions={roomData.iframeOptions}></IFrameRoom>,
-            adventure: <Adventure options={roomData} onClick={this.onAdventureClick.bind(this)}></Adventure>
+            adventure: <Adventure options={roomData} onClick={this.onAdventureClick.bind(this)}></Adventure>,
+            library: <LibraryRoom libOptions={roomData.libOptions}></LibraryRoom>
```
morganecf, line 89: I would call this `ImageMapRoom` or something more generic, since I can see us using this sort of setup for like, an art gallery.

app/client/jsx/LibraryRoom.jsx
```diff
@@ -0,0 +1,33 @@
+import React, { Component } from 'react'
+
+class LibraryRoom extends Component {
```
morganecf, line 3: We've been trying to switch to using functional components. You can rewrite this pretty easily as a function. See `Poke.jsx` for an example

app/client/jsx/Room.jsx
bMacSwigg: Done. I changed this and similar to ImageMapRoom, and made other changes to be less specific to libraries (e.g. "category" -> "label")

app/client/jsx/LibraryRoom.jsx
bMacSwigg: Done (I think I did that correctly? It works, anyhow).

morganecf: Looks good!

bMacSwigg:
A bit rough around the edges (especially with the styling), but the main functionality is here now.

bMacSwigg:
main
after-topic-select
after-pdf-select

app/client/jsx/ImageMapContents.jsx
```diff
@@ -0,0 +1,62 @@
...
+        </div>
+    )
+}
+
+export default props => {
+    const [itemIsShown, setItemIsShown] = useState(false)
+    const [item, setItem] = useState({})
+    if (!itemIsShown) {
+        return displayList(props, onSelectItem(setItemIsShown, setItem))
+    } else {
```
morganecf, line 59: This `else` is redundant

app/client/jsx/ImageMapContents.jsx
```diff
@@ -0,0 +1,62 @@
...
+                <button onClick={returnFunc}>
+                    <FontAwesomeIcon icon={faArrowLeft}/>
+                </button>
+                <b>{item.title}</b>
+                <button onClick={props.handleClose}>
+                    <FontAwesomeIcon icon={faTimes}/>
+                </button>
+            </div>
+            <div className="contents">
+                <PDFObject url={item.path} />
```
morganecf, line 48: Does this pdf component allow us to show images as well? I'm not really sure how PDF viewers work

app/client/jsx/ImageMapContents.jsx
bMacSwigg: Good point; removed.

app/client/jsx/ImageMapContents.jsx
bMacSwigg: I assume you mean, allow `contents` elements that are .jpg or .png in addition to .pdf? (As opposed to showing a preview image of a PDF as mentioned above).

I don't think it supports images natively, but we could add logic around the path to detect the filetype and display it in different ways.

bMacSwigg:
Thanks! Responses inline.

> 1. You might already be on this, but we definitely need to have the click areas highlighted on hover

I've spent some time looking into that and plan to have something for it by Saturday. There are libraries that will do it, but so far I've mostly found jQuery libraries; I'd like to find something more tailored to React.

> 2. Clicking outside the modal / hitting escape doesn't let me escape the modal

Should be fixed now.

> 3. How hard would it be to show a preview of each PDF instead of listing out links? Not necessary for now, but if it's easy maybe we can copy the google drive design for a better experience

Might be tricky. The PDF viewer doesn't natively support it. It might be possible with Mozilla's PDF.js, but I'm not sure how much custom logic it would require. I can look into this more once the bulk of the feature is done.

> 4. Same question for toggling through the PDFs instead of going back to the list to view a new one. Not sure if keybindings would interfere with the ones on Room.jsx

This doesn't seem terribly difficult, but I'll similarly wait to work on this until the main feature has been nailed down.

morganecf:
@bMacSwigg got it. Can you merge this for now, once conflicts are resolved, and then open those last two changes in a separate PR (time permitting)?

PR #152, opened 2020-07-22
After sign-in navigate to room specified as first room in config

morganecf:
Also remove an old file


app/client/jsx/Welcome.jsx
```diff
@@ -77,4 +77,3 @@ export default connect(
     {
         addConfig: reducers.addConfigActionCreator,
      })(Welcome)
-
```
morganecf, line 80: There is a new line at the end of this file…I'm confused

PR #162, opened 2020-07-24
TST Event config

morganecf:
Mostly just config, and I also commented out a bunch of Puck Select stuff to make it open as a grid instead of by row FYI @danielmdavis


app/client/jsx/IFrameRoom.jsx
```diff
@@ -2,10 +2,22 @@ import React from 'react'
...
+                    width="100%"
+                    frameBorder="0px"
+                >
+                </iframe>
+            </div>
+    ) : <JitsiVideo jitsiData={props.jitsiData}></JitsiVideo>
     return (
         <div className="iframe-room">
-            <JitsiVideo jitsiData={props.jitsiData}></JitsiVideo>
+            {left}
```
morganecf, line 20: This was so that I could put two iframes in one room. This is a hack for now, but I want to be able to extend the config to take multiple iframes -- or even better, change the way we do rooms so that they're arbitrarily composable with iframes, video, chat, content, etc.

PR #163, opened 2020-07-24
Fix bug wherein no room is found

seangibat:


seangibat:
Note that I didn't test this whatsoever and do not plan to

PR #169, opened 2020-07-25
Full-page and resizable image map

bMacSwigg:
Uses the div size to calculate the maximum size that can fit and sets it on the react component.

Somewhat tricky because image maps don't natively support resizing (the area coordinates are hard-coded pixel values). The react component can handle it, but as a result we lose the benefit of having CSS automatically size the image. Instead we have to manually calculate the maximum width in pixels to set it pass it to the react component.

Since that width is dependent on the size of the surrounding div, which doesn't exist until the React component is rendered, we also have to useEffect() to calculate the size (and then do some checks to make sure we don't keep re-rendering infinitely).

But, the end result is a nice (fairly) smoothly resizing image map! :)


bMacSwigg:
Demo of resizing, fitting to both the height and the width of the screen, and how the relevant area stays consistent.

ImageMap size 1
ImageMap size 2

PR #170, opened 2020-07-25
Record user IP (continuation of #112)

bMacSwigg:
This finishes the open #112 to be proxy-aware and configurable.

Tested both with & without proxies (with the help of @hndrewaall )


hndrewaall:
Haven't actually tested, but will be easier to test from master

PR #176, opened 2020-08-08
Proto schemas for JSON configs

bMacSwigg:
This PR adds new `.proto` files that define the structure of the various JSON config files. `config.py` uses the proto files to enforce structure on the JSON files, but otherwise we do not transport or store anything in proto format. This PR also includes some changes to the structure of the JSON configs to make them compatible with protobufs; these changes are largely cosmetic.

### Reason for proto schemas

Proto schemas allow us to give self-documenting and self-enforcing structure to the json configs. As it is without this change, the only way to know what is possible in the configs is to:
- Look at the examples that exist, and
- Trace through code to see what is actually used

Both of these options are time-consuming, not beginner-friendly, and error prone.

With proto schemas, you can look at the schema to see the complete list of values accepted, accompanied by a brief comment explaining the field. Because we validate using the schema, it is impossible for new values to be used without also updating the proto, ensuring the schema stays up to date.

### Constraints of the proto format

There are two main ways in which JSON is more flexible than proto, which conflicted in minor ways with our existing configs:
1. JSON allows arbitrary dict keys; proto does not. (This is in essence what it means for the proto to have a schema, anyways. This is unavoidable.) This means that our existing set up for configs like `rooms.json`, where we have an arbitrary room ID as the key to the room object itself, isn't supported.
2. JSON allows 2D arrays; proto does not. This is unfortunate, but not a big complication. In proto instead we have an array of objects, where each object contains another array as a field. This only affects the avatar sets.

### Changes to current configs

To address these and other minor compatibility issues, I have made the following changes to the existing configs:
- Because of (1) above, the rooms cannot be keyed off of their ID in JSON, as mentioned. However, being able to use the set of rooms as a dict rather than as a list is very useful. I've changed the JSON representation to be a list where each room contains a new `id` field, but in `config.py` I translate it back into a dict keyed off of the ID, so the React code can continue using it as a dict like before.
- Also because of (1) above, the avatar sets cannot have string keys. The keys weren't being used for anything other than opaque references into the sets, so it didn't seem worthwhile to do the same translation back into dicts. Instead, I've changed all avatar-related code to use the indices directly instead. This happily ended up simplifying the avatar code along the way.
- Because of (2) above, the avatar sets also cannot be a simple 2d array. To access an avatar, we have to do `avatars[firstIndex].images[secondIndex]` instead of just `avatars[firstIndex][secondIndex]`.
- `RoomType`, as an enum, has to be represented in ALL_CAPS. I've changed this in JSON and wherever room types were referenced explicitly in `.jsx` files.
- Because of (1) above, adventures similarly can no longer have arbitrary IDs as keys. For the top-level adventure objects, those keys are actually completely unused, so translating it directly into a list is almost a no-op. For the adventure rooms themselves, I take the same approach as in `rooms.json` of translating back into a dict from the list and `id` fields.
- In the adventure rooms, `nextRoom` needs to be specifically typed, so I've made it always a list now. This is a no-op change on the React side, since "choose a random element out of 1 element" is the same as "if it's not a list, go to the single element, otherwise random".
- Again, because of (1) above, imagemaps need to have `id` fields. I do the same translation to dict.

### Additional changes

This PR also includes some additional changes not in the proto files or described above.
- Adds dependency on protobuf in pip and with brew/apt (documented in README).
- Adds a Makefile for building the protos.
- Restructures `config.py` to load & validate each set of configs separately, since the files now have associated types. Also improves error messages.


app/config/schema/config.proto
```diff
@@ -0,0 +1,216 @@
...
+ * top-level configuration (base.json, development.json, production.json)
+ *
+ * Proto doc: https://developers.google.com/protocol-buffers/
+ */
+
+syntax = "proto3"; // Must be proto3 to support JSON
+
+message Config {
+    // TODO: What does this flag do?
+    bool debug = 1;
```
bMacSwigg, line 12: @morganecf @hndrewaall What all is the debug flag for? I saw that it automatically turns on some of the navigation bar stuff (map etc), but does it do other stuff too?

app/config/schema/config.proto
```diff
@@ -0,0 +1,216 @@
...
+    // Number of server-side proxies in use (if unsure, probably 0)
+    // Required.
+    int32 num_proxies = 14;
+}
+
+message Moderation {
+    // Contact emails for moderators to receive moderation form submissions
+    // TODO: Is this currently used?
+    // Required (at least one).
+    repeated string moderator_emails = 1;
```
bMacSwigg, line 80: @morganecf @hndrewaall it seems like nothing currently calls the send_email endpoint. Is that accurate? Is that temporary, or has email been deprecated in favor of text? How should I document this?

app/config/schema/config.proto
```diff
@@ -0,0 +1,216 @@
...
+    AvatarSet goodbye = 1;
+}
+
+message MapConfig {
+    // Options related to the stats to display above the map
+    // Required.
+    MapStats stats = 1;
+
+    // TODO: What is this, and what values can it take?
+    string scale_by = 2;
```
bMacSwigg, line 143: @morganecf @hndrewaall help <3

app/config/schema/rooms.proto
```diff
@@ -0,0 +1,171 @@
...
+    string description = 4;
+
+    // Relative filepath to an image to display behind room content
+    // Optional
+    string background_image = 5;
+
+    // Color for the entrance door, formatted as a hex code or HTML color name
+    // TODO: What is the default?
+    // Optional
+    string door_tint = 6;
```
bMacSwigg, line 49: @morganecf @hndrewaall What's the default value for this? (I'm sure I could've figured this out on my own, but also, I was feeling lazy and I'm tired)

app/config/schema/rooms.proto
```diff
@@ -0,0 +1,171 @@
...
+
+    // ID of the room to the west
+    // Optional
+    string west = 4;
+}
+
+message MapDisplay {
+    // TODO: is this accurate?
+    // X and Y coordinates of the top-left corner of the room
+    // (0,0) is the top-left of the map
```
bMacSwigg, line 116: @morganecf @hndrewaall Am I correct to say these are both top-left here? (That 0,0 is top-left of the map, and this x,y is top-left of the room relative to that)?

app/config/schema/config.proto
morganecf: Using text instead of email was a super last minute change (as was doing the whole thing via a subprocess call). Let's keep this documented for now, because we'll want to switch back to email. We also want to keep emails out of the config and set them as env variables or something, since it'd be better not to commit them.

app/config/schema/rooms.proto
morganecf: That is correct!

app/config/schema/rooms.proto
morganecf: [There doesn't appear to be one.](https://github.com/morganecf/jitsi-party/blob/master/app/client/jsx/Door.jsx#L38) Let's just set it to black as a default.

app/config/schema/config.proto
morganecf: I think…it was intended to be more of a "developer mode" flag. I don't think it does anything else other than turn the map on automatically so that devs have access to it. It seems useful but I forgot it existed!

app/config/schema/config.proto
morganecf: This was also a last-minute pre-party hack…it's supposed to control whether or not the map room colors are scaled by opacity or by a d3 color scale, but I think I ended up just panic-commenting the color scale logic out of `MapVisualization.jsx`, so it's not even used. Let's get rid of it from the config!

app/config/schema/rooms.proto
bMacSwigg: Great! Set this to black and removed TODO comment.

app/config/schema/config.proto
bMacSwigg: Cool, removed this

app/config/schema/config.proto
bMacSwigg: I added an initial comment on this. We can adjust as needed.

app/config.py
```diff
@@ -9,12 +12,57 @@ def get_json_dict(path):
     try:
         with open(path) as file:
             return json.load(file)
-    except:
+    except Exception as e:
+        print("Error on path ", path, e)
```
morganecf, line 16: Flask commands now complain about a missing `config/events.json` file. I think we might want to add a blank `events.json` to avoid this error.

app/manager.py
```diff
@@ -41,8 +41,8 @@ def create_db():
     db.create_all()
 
     # Load room definition
-    roomConfigs = [os.path.join(folder, 'rooms.json') for folder in [configdir, overridedir]]
-    roomConfig = make_merged_cfg(roomConfigs)['rooms']
+    roomConfigPaths = [os.path.join(folder, 'rooms.json') for folder in [configdir, overridedir]]
+    roomConfig = load_and_validate_rooms(roomConfigPaths)
```
morganecf, line 45: It occurred to me that this might be redundant, since [the application context is initialized inside `manager.py`](https://github.com/morganecf/jitsi-party/blob/master/app/manager.py#L12). This instantiates a config object, whose constructor already calls `load_and_validate_rooms`. We should be able to reference `app.config['ROOMS']` to populate the db with rooms, instead of calling `load_and_validate_rooms` for the second time.

bMacSwigg: You're right! Just tested it and it seems to work fine. Once I ran `flask create-db` I could query all the rooms from the flask shell, and when I ran & connected to the app it all seemed to work as expected.

app/config/schema/rooms.proto
```diff
@@ -0,0 +1,169 @@
...
+    // Relative filepath to an image to display behind room content
+    // Optional
+    string background_image = 5;
+
+    // Color for the entrance door, formatted as a hex code or HTML color name
+    // Optional. Defaults to black.
+    string door_tint = 6;
+
+    // Maximum number of people who can be in the room at once
+    // If not specified, defaults to infinite
```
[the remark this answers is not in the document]
morganecf: Yeah, on second thought, 10 is only an upper bound if people have their vide on, and we have the ability to force-disable video.

app/manager.py
morganecf: Cool!

app/config.py
morganecf: Nice, thanks!

PR #183, opened 2020-08-28
Support event open/close times

bMacSwigg:
This adds a config option for open/close times. This will not boot users after the close time, but it will prevent users from joining before or after an event.

Screenshot of display outside of event times:
Screenshot 2020-08-27 at 9 08 37 PM

### DIFF
--- a/app/client/jsx/Welcome.jsx
+++ b/app/client/jsx/Welcome.jsx
@@ -58,7 +58,31 @@ class Welcome extends Component {
         this.setState({ authenticated: true })
     }
 
+    isOpen() {
+        if (!Config.eventTimes) return true;
+
+        const now = new Date()
+        if (Config.eventTimes.start) {
+            const start = Date.parse(Config.eventTimes.start)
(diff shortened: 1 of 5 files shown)


app/config/overrides/config.json
```diff
@@ -92,5 +92,10 @@
         "unlockText": "You have unlocked the ability to hex",
         "selfPoke": "Ouch",
         "unlockAfterMinutes": 0
+    },
+    "eventTimes": {
+        "start": "2020-08-27T19:00:00-04:00",
+        "end": "2020-08-27T01:00:00-04:00",
+        "closedText": "Sorry, the event is closed"
```
SeanMcGrath, line 99: proto says `closed_html`

SeanMcGrath (commented):
just one thing, otherwise LGTM

app/config/overrides/config.json
bMacSwigg: Oop. Should be fixed now.

PR #185, opened 2020-08-28
Improve config management, add Gojo v1

SeanMcGrath:
An attempt to normalize/modularize configuration of the various app installations.

Basic idea is that the override configs are no longer real files, they are symlinks - the repo ships with all the configs a user might want, and you just symlink all the custom config files into the overrides folder before starting flask/webpack.

Also added super basic gojo config.

In testing, was able to switch between default/gojo/TST configs by doing e.g.
```
rm config/overrides/*
ln -s `pwd`/config/gojo-video/* overrides/
flask create-db
flask run
```


app/client/jsx/JitsiVideo.jsx
```diff
@@ -107,6 +107,7 @@ class JitsiVideo extends Component {
             }
 
             window.api = new window.JitsiMeetExternalAPI(domain, options)
+            debugger
```
bMacSwigg, line 110: Remove? (elsewhere too)

app/config/README.md
```diff
@@ -2,15 +2,15 @@
 
 This folder contains the default configuration files for the jitsi-party application.
 Both the server and the client use these files.
-Do not alter them directly.
 
-To configure your custom installation of jitsi-party, drop new files into the `overrides` folder here.
+`base` contains the basic/default configurations, on which oyur custom configs will build. Do not alter these files.
```
bMacSwigg, line 6: Nit: typo

app/config/README.md
```diff
@@ -2,15 +2,15 @@
...
+
+To configure your custom installation of jitsi-party, symlink your configuration files into the `overrides` folder.
+The contents of this folder are gitignored, so the symlinks will not be stored in the repository.
+By convention, a party name `my-party` will keep its configs in `config/my-party/`, and you can symlink `config/my-party/*` into `config/overrides` when you are ready to run your party server.
```
bMacSwigg, line 10: I know the setup process to symlink is already fairly simple, but I wonder if it's worth adding a script that just takes the party name as a param. So `setup_configs my-party` clears out the existing overrides and symlinks in `config/my-party`.

app/manager.py
```diff
@@ -5,7 +5,7 @@
 from config import load_and_validate_rooms
 
 basedir = os.path.abspath(os.path.dirname(__file__))
-configdir = os.path.join(basedir, 'config')
+configdir = os.path.join(basedir, 'config', 'base')
 overridedir = os.path.join(configdir, 'overrides')
```
bMacSwigg, line 9: This needs to be updated too now -- probably easiest to just do `os.path.join(basedir, 'config', 'overrides')`?

app/config.py
```diff
@@ -6,6 +6,7 @@
 
 basedir = os.path.abspath(os.path.dirname(__file__))
 configdir = os.path.join(basedir, 'config')
```
bMacSwigg, line 8: Can we get rid of this now, and just construct the basecfg & override dirs off of `basedir`?

app/config/gojo-video/rooms.json
```diff
@@ -0,0 +1,46 @@
+{
```
bMacSwigg, line 1: Some weird indentation in this file mixing tabs and spaces. I'm in favor of using spaces for everything and getting rid of tabs, personally, but it would be good to fix this one way or the other.

app/config/README.md
SeanMcGrath: down to do this as a separate change

PR #188, opened 2020-09-03
Make page title configurable, make index a template

SeanMcGrath:


.gitignore
```diff
@@ -163,3 +163,6 @@ dist
 
 # Generated python files for proto schemas
 app/jitsi/schema/*pb2.py
+
+# VSCode
+*.code-workspace
```
bMacSwigg, line 168: Nit: newline at end of file

app/client/styles/themes/_active.scss
```diff
@@ -0,0 +1 @@
+/Users/smcgrath/code/jitsi-party/app/client/styles/themes/_tst.scss
```
bMacSwigg, line 1: Can this be a relative path?

app/config/schema/config.proto
```diff
@@ -13,9 +13,14 @@ message Config {
     // Optional. Defaults to false.
     bool debug = 1;
 
+    // Title for main page
+    // document.title in the DOM
+    // Required.
+    string pageTitle = 18;
```
bMacSwigg, line 19: Super nit: this should technically be `page_title` according to proto style. The json parser will still accept `pageTitle` in the config files.

bMacSwigg (approved):

app/client/styles/themes/_active.scss
SeanMcGrath: oh this should be gitignored

PR #191, opened 2020-09-04
allow selection of theme via symlink

SeanMcGrath:
decided it would be a good idea to handle themes the same way we handle config


app/client/styles/main.scss
```diff
@@ -13,10 +13,9 @@
 
 // 3. Themes
 @import
-  // 'themes/default',
-  // 'themes/june13';
-  // 'themes/bauhaus';
-  'themes/tst';
+// symlink your theme file to "themes/_active.scss" and it will override the default
+// if you don't have a theme, just `ln -s _default.scss _active.scss`
+  'active';
```
bMacSwigg, line 18: Wouldn't this need to be 'themes/active'?

bMacSwigg (approved):

app/client/styles/main.scss
SeanMcGrath: no, the change to the webpack config makes the prefix unnecessary. instead, we resolve `themes/_active.scss` preferentially over `themes/default/_active.scss`

PR #193, opened 2020-09-04
Basic markdown support

bMacSwigg:
Basic markdown in room and event descriptions. Supports:
- _emphasis_ (`*text*` or `_text_`)
- **strong** (`**text**` or `__text__`)
- ~delete~ or strikethrough (`~text~`)
- line breaks
- [links](http://www.google.com) (`[text](http://www.google.com)`)


app/client/jsx/Room.jsx
```diff
@@ -95,7 +96,11 @@ class Room extends Component {
         if (this.props.rooms[this.state.room].description) {
             return (
                 <div className="room-content">
-                    {this.props.rooms[this.state.room].description}
+                    <ReactMarkdown
+                        source={this.props.rooms[this.state.room].description}
+                        allowedTypes={['text', 'break', 'emphasis', 'strong', 'delete', 'link']}
```
SeanMcGrath, line 101: ignorable nit: i'd pull this list out to a constant

SeanMcGrath (approved):

app/client/jsx/Room.jsx
bMacSwigg: (For future reference): we discussed; agreed that this would be lovely but that there's not an immediately obvious place to put this.

PR #203, opened 2020-10-09
Dockerized local development

hndrewaall:
Currently broken by jitsi component dropping `/jitsi` prefix from its calls to the jitsi-meet backend (`/http-bindroom` etc). If anyone can help would greatly appreciate!


hndrewaall:
Note that this is not based on latest master, nor do I wish to rebase until I can figure out wtf is going on here

hndrewaall:
Also this is very much a WIP so there's def some noise in here (like my pyenv crap)

hndrewaall:
OK ready!

hndrewaall:
@morganecf i think maybe you dropped the "basic" theme, which was previously "active" (which is now a gitignored symlink). so this adds it back. probably also needs the fix you did for the others?

hndrewaall:
@morganecf actually the "basic" css/theme combo works but looks pretty weird to me. i wonder if tst specific stuff is being leaked somehow

morganecf:
@hndrewaall I didn't drop `_basic.scss` : https://github.com/morganecf/jitsi-party/commit/44bbf6a76946dd7ca57130284c9efa2746963048 

The important thing I did in that commit was to move `_active.scss` to the `themes` directory. I explained this in the PR, but to be more clear, there were 2 issues:
1. Old themes are missing variables that got added to other stylesheets during TST development. The TST themes set these variables, so you shouldn't see any issues like `Undefined variable: "$avatar-highlight-color".` for those once 2) is fixed. **This isn't fixed.**
2. Because `_active.scss` was in the wrong place, setting the theme wasn't really working and the main stylesheet was importing an old theme, triggering error 1). **This is fixed**, meaning all the TST themes should be working.

hndrewaall:
yeah i figured out the move bit. still though it doesn't look right. has tst fonts and stuff

morganecf:
I don't really know what `basic` is or where it came from, but I agree that it shouldn't have TST fonts. Maybe we linked a TST theme to `basic` at some point?

hndrewaall:
no, I'm renaming active to basic, because active connotes that it is active, which it isn't

hndrewaall:
anyway i think i figured it out

morganecf:
I'm confused -- aren't we using `active` as the active theme? It gets imported in `main`. How are we setting the theme otherwise?

hndrewaall:
don't worry about it. it's fixed now. active is a symlink. the old active css is now called basic

hndrewaall:
however, said basic css still results in tst fonts which seems wrong

morganecf:
Ok I see. The old `active` CSS definitely contained TST styles because I hardlinked the TST theme to it.

hndrewaall:
OK. we used to have a non TST theme that i would really like to have back

hndrewaall:
we need to not break everything non-TST every time we add something for them

hndrewaall:
aside from weird fonts with the default css, things work now here and are ready for review

morganecf:
I'm poking around in history and there used to be a theme called `_default.scss`, but I'm not sure where it went. We might have overwritten it at some point with the `default/_active.scss` business. I agree that we shouldn't break things, obviously…I made a ticket for the inconsistent theme issue: https://github.com/morganecf/jitsi-party/issues/322

README.md
```diff
@@ -4,49 +4,15 @@ A virtual party space.
...
-# Set up a virtualenv
-virtualenv venv
-source venv/bin/activate
-pip install -r requirements.txt
-
-# Make sure you have node/npm installed
-brew install node # on Linux: `apt install npm`
-
-# Make sure you have protoc installed
-brew install protobuf # on Linux: `apt install protobuf-compiler`
```
bMacSwigg, line 18: Do users need to do anything to get these dependencies now (node and protoc)?

app/setup.sh
```diff
@@ -1,15 +0,0 @@
-export FLASK_APP="manager.py"
-export FLASK_ENV=development
-export FLASK_RUN_PORT=3000
-
-if [ -f email_config.sh ]; then
```
bMacSwigg, line 5: Do we still use these email_ and text_config files?

bMacSwigg (commented):
This generally LGTM, but I haven't had the time to actually try running this on a GCP VM. The documentation seems good and I very much appreciate that setup is consolidated into the Makefile more now. I left a couple comments about things that were deleted.

README.md
hndrewaall: This should stay. I'll restore

app/setup.sh
hndrewaall: I need to figure out what prod is doing and update this accordingly. probably it'll just be stuff to add to the .env file but that should be documented

tstsocal:
A couple of issues I noticed:

1. In the **gen-passwords.sh** it doesn't like the empty quotes after the `sed -i`.
2. In the **env.template.sh** you need to remove white-space on lines: 55 & 56, should be:
         BRIDGE_CHANNEL=datachannel
         ENABLE_XMPP_WEBSOCKET=false

tstsocal:
A couple of issues I noticed:

1. In the **gen-passwords.sh** it doesn't like the empty quotes after the `sed -i`.
2. In the **env.template.sh** you need to remove white-space on lines: 55 & 56, should be:
         BRIDGE_CHANNEL=datachannel
         ENABLE_XMPP_WEBSOCKET=false

hndrewaall:
Updated docs and rebased and force-pushed. Gonna executive merge this for everyone's sanity

PR #207, opened 2020-10-10
Update config.json

kacycrider:
Removed the hatewatching stat, changed the rooms for meeting animals to reflect where animals are, changed the header to say welcome home, changed the video display name from Fellow Satanist to Satanist.

### DIFF
--- a/app/config/tstvhq/config.json
+++ b/app/config/tstvhq/config.json
@@ -69,7 +69,6 @@
                 "text": "📍 {count} exploring"
             },
             "additionalStats": [{
-                {
                     "text": "📕 {count} reading",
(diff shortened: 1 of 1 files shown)


app/config/tstvhq/config.json
```diff
@@ -68,17 +68,13 @@
                 "display": false,
                 "text": "📍 {count} exploring"
             },
-            "additionalStats": [{
```
hndrewaall, line 71: this shouldn't be deleted i don't think

hndrewaall (changes requested):

hndrewaall (approved):

PR #214, opened 2020-10-16
Update rooms 10-16_2.json

LEDFlashing:


app/config/tstvhq/rooms.json
```diff
@@ -127,10 +127,13 @@
         {
             "id": "teaRoom",
             "name": "Tea Room",
-            "type": "JITSI",
-            "capacity": 10,
-            "muteRoom": false,
+            "type": "iFrame",
```
hndrewaall, line 130: this might be case sensitive?... we'll find out!

hndrewaall (approved):
approving but might need a fix. will test after merging cuz fuck it

PR #270, opened 2020-12-21
Update rooms.json

kacycrider:
Updated infernal theater and barnyard links

### DIFF
--- a/app/config/tstvhq/rooms.json
+++ b/app/config/tstvhq/rooms.json
@@ -165,7 +165,7 @@
             "name": "Barnyard",
             "type": "IFRAME",
             "iframeOptions": {
-                "src": "https://player.twitch.tv/?channel=disasterina&parent=estate.tstvhq.com"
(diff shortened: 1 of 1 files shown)


app/config/tstvhq/rooms.json
```diff
@@ -165,7 +165,7 @@
             "name": "Barnyard",
             "type": "IFRAME",
             "iframeOptions": {
-                "src": "https://player.twitch.tv/?channel=disasterina&parent=estate.tstvhq.com"
+                "src": "https://youtu.be/N_DwST8vKdQ"
```
hndrewaall, line 168: I don't think shortened urls are going to work

hndrewaall (changes requested):
need new url

app/config/tstvhq/rooms.json
kacycrider: ah! thanks. its cool, we can skip this-- was just placeholder

hndrewaall (approved):

PR #329, opened 2021-01-31
Update adventures.json

kacycrider:
how did i do at braces


app/config/tstvhq/adventures.json
```diff
@@ -85,41 +85,4 @@
...
-                    },
-                    {
-                        "text": "Twenty Four Please",
-                        "nextRoom": ["twentyFour"]
-                    }
-                ]
-            }
-        ]
         }
-    ]
```
hndrewaall, line 124: this deletion looks wrong to me. did you try running this locally?

kacycrider: This was @LEDFlashing -- Jack, do you need to make changes? Unclear from Andrew's comment.

Andrew, no, I haven't been able to run anything locally since the transition to VS Code.

LEDFlashing: I certainly didn't. if you're saying I need to undo the deletion I'll just do it manually.

LEDFlashing: err ... if I had edit access to that one.

hndrewaall:
closed automatically due to history rewrite. can reopen when pr branch is rebased (but i dont think we need this anyway)

PR #360, opened 2021-02-15
Update events.json

LEDFlashing:


hndrewaall (changes requested):
bats and wolves need update i think

hndrewaall (approved):

PR #385, opened 2021-03-23
Adding Room type with Converse.js replacing jitsi

rene-grigory:
# Issue
Jitsi Rooms using the jitsi-meet for chat only consume a large quantity of processing resources making the rooms unusable at around 75 users.

# Solution
Created a room type called CHATSTREAM. CHATSTREAM is very similar to an IFRAME room type except it does not make use of JitsiVideo. Instead, it uses a third party library called [Converse.js](https://conversejs.org/). 
- The minified code for Converse.js was downloaded and installing in the _app/client/js_ folder
- The CSS is pulled in from a CDN (we could download this locally, but there was no noticeable blocking pulling from the CDN)
  -  Note: the original code uses SASS and Bootstrap for generating the CSS. Rather than pull all of the required Bootstrap and SASS folders from the source I decided it would be better to create an _app/client/styles/_chatStreamRoom.scss_ to override the necessary styles.
  - To set the color for the theme the SASS uses the **$backgound-color** variable

# Additional Comments
- I noticed that the _app/client/node_modules_ was not in the _.gitignore file. I added it; we can undo if desired. 
- Converse.js injects the DOM Element at the bottom of the body so I made two changes to deal with this:
  1. I moved the **<script />** tags into the **<head />** and applied the **defer** attribute. This causes the scripts to loaded while the DOM is being parsed and then executes the scripts after parsing, but prior to rendering in the order the scripts were called
  2. I used **absolute** positioning to force the Converse DOM Element into a "reasonable" position on the page
      - I also kept a **div** with the **jitsi-video** class in the page to keep the left side opened for the Converse DOM Element to appear in
  3. There is a log going on in the **_useEffect()_** method. We can Zoom or chat as needed.


hndrewaall:
This looks really good! I'll check it out locally and play around with it. @morganecf could you take a look as well?

hndrewaall:
For everyone's benefit, here's a screenshot from my local env:
image

hndrewaall:
With this is it possible to control chat colors from the theme?

hndrewaall:
image
I triggered this somehow

hndrewaall:
image
Also it seems users aren't cleaned up. Is there a timeout we can control?

hndrewaall:
Also this is pretty minor, but the rooms don't seem to be listed in the chat server. I wonder if there's a setting that controls that in the client

hndrewaall:
Timeouts aside, I wonder if it's possible to hook room leave/window close/other things to explicitly have users leave rooms

hndrewaall:
Also do you have any idea what the "hidden message" thing is supposed to do?

hndrewaall:
BTW so this is what happens when you join with the same name:
image

I can see both pros and cons to this approach

rene-grigory:
Some responses:

**With this is it possible to control chat colors from the theme?** I'm currently using the $background-color variable to set the borders/icons, but if we want to change other colors we'll need to identify which areas on the screen and which SASS variables to use.

**I triggered this somehow** No clue how you did that or what it even means.

**Also it seems users aren't cleaned up. Is there a timeout we can control?** It seems that Prosody requires that each room has an "owner". It assigns that role to the first user to join the room. If the "owner" leaves the room then Prosody generates a fake "owner": <uuid>@guest.party.jitsi. I have no idea how to deal with that. 

**Also this is pretty minor, but the rooms don't seem to be listed in the chat server (as in, from the CLI client).** Sorry, no clue on my part.

**Timeouts aside, I wonder if it's possible to hook room leave/window close/other things to explicitly have users leave rooms** I should be able to do that by listening to the onBeforeUnload (actually, looks like react-beforeunload is a dep so I can use that).

**Also do you have any idea what the "hidden message" thing is supposed to do?** I'm not sure what "hidden message" are you talking about.

**BTW so this is what happens when you join with the same name:** As far as the multiple nicks: since we don't force unique aliases I can see how having 10 Omens in a chat might get confusing. Allowing the users to alter their name for a given chat seems like a good idea to me. FYI: this is being enforced by Converse/Prosody; not me.

rene-grigory:
I've added the beforeUnload listener to clean out the user.

hndrewaall:
I see some settings here that may be relevant to things i mentioned: https://prosody.im/doc/modules/mod_muc

I will try playing around on the backend!

rene-grigory:
So, I see that the "owner" is treated as if it is offline. In fact, when I have multiple users in the room and one leaves (logs out), they still show as offline. I came across some initialization settings for the UI:
```
hide_offline_users: true,
muc_fetch_members: false,
```
These setting do prevent the standard offline members from showing in the list, however, the "owner" still shows. If I could hide that I would feel better. I'll look into it some more this evening.

hndrewaall:
I think we might be able to disable the requirement for owners sticking around. I'm specifically looking at that "persist" option in the docs I linked

hndrewaall:
BTW I am seeing a few console errors from the chat. Here's one I seem to see every time I load the chat: 
image

hndrewaall:
image
Another

hndrewaall:
image
BTW this is the "Hidden Message" feature I mentioned. I don't know what it's supposed to do, but I don't think it's working

rene-grigory:
It's a rather sill feature. See my screens shots below:
image
image
image

morganecf:
Thanks for this @rene-grigory! I'll take a look tomorrow :)

hndrewaall:
> It's a rather sill feature. See my screens shots below:

okay this is basically what i expected. weird that it doesn't seem to do that in my env. i wonder what's broken

rene-grigory:
I went to their site and opened a demo and I see the same error in the console that @hndrewaall reported:

image

app/client/jsx/ChatStreamRoom.jsx
```diff
@@ -0,0 +1,84 @@
...
+
+    // configure the 'jitsi-plugin' to get our hands on the converse api logout
+    try {
+      window.converse.plugins.add("jitsi-plugin", {
+        initialize: function () {
+          logout = this._converse.api.user.logout;
+          plugins = this._converse.pluggable.plugins;
+
+          // add event listener to restucture the chatbox
+          this._converse.api.listen.on("chatBoxInsertedIntoDOM", () => {
```
morganecf, line 34: Apparently the `chatBoxInsertedIntoDOM` event will be removed in the next major version: https://github.com/conversejs/converse.js/blob/master/CHANGES.md

We could replace it with one of these events: https://github.com/conversejs/converse.js/blob/master/src/plugins/minimize/index.js#L128-L133

app/client/jsx/Room.jsx
```diff
@@ -1,275 +1,314 @@
...
+import IFrameRoom from "./IFrameRoom.jsx";
+import ImageMapRoom from "./ImageMapRoom.jsx";
+import Door from "./Door.jsx";
+import Adventure from "./Adventure.jsx";
+import Navigation from "./Navigation.jsx";
+import { PokeNotification } from "./Poke.jsx";
+import { WebSocketApi } from "./WebAPI.jsx";
+import LocalStorage from "./LocalStorage.jsx";
+import Config from "./Config.jsx";
+import { ChatStreamRoom } from "./ChatStreamRoom.jsx";
```
morganecf, line 21: The diff in this file is a little confusing because of what I presume is linting…is there any chance you could commit without the linting? That way our history will be cleaner and reviews will be easier!

app/client/jsx/Room.jsx
```diff
@@ -1,275 +1,314 @@
...
+        <div className="room-content">
+          <ReactMarkdown
+            source={this.props.rooms[this.state.room].description}
+            allowedTypes={[
+              "text",
+              "break",
+              "emphasis",
+              "strong",
+              "delete",
+              "link",
```
morganecf, line 126: Nice! This is great. I test it out just to see and it looks good: 
Screen Shot 2021-03-25 at 5 31 34 PM

app/config/schema/rooms.proto
```diff
@@ -93,6 +94,11 @@ message Room {
     // The page to redirect to
     // Required for redirect rooms.
     string route = 16;
+
+    // CHATSTREAM ROOMS ONLY
+    // Options related to the bosh service
+    // Required for chat streaming rooms.
+    string boshUrl = 17;
```
morganecf, line 101: Should this be in `Room`? Since `boshUrl` is always the same, does it need to be part of the room config? Can we make it part of the general config?

morganecf (commented):
I played around with it for a bit and it looks awesome!

I was wondering whether we could control the user's converse chat profile picture, to stay consistent with the rest of the app. I was poking around the code and it seems like we can use the `vcard` API to update profile information: https://github.com/conversejs/converse.js/blob/master/src/headless/plugins/vcard.js#L291-L315

It expects a base64 encoded image, so we'd have to download and encode the jitsi user's `avatar`, which can be accessed here: https://github.com/morganecf/jitsi-party/blob/master/app/client/jsx/Room.jsx#L82 

I ended up getting a proof of concept working with the following, where the image is just a square with a diffusion of red in it:

```javascript
this._converse.api.listen.on("VCardsInitialized", () => {
            this._converse.api.vcard.set(this._converse.bare_jid, {
              image: 'iVBORw0KGgoAAAANSUhEUgAAAAUAAAAFCAYAAACNbyblAAAAHElEQVQI12P4//8/w38GIAXDIBKE0DHxgljNBAAO9TXL0Y4OHwAAAABJRU5ErkJggg=='
            })
})
```

This doesn't have to be done in this PR, of course! I figured I'd share since I ended up following the question down a rabbit hole. 

EDIT - I realized this may have been unclear so here's a screenshot of what I mean, where the user's default avatar is replaced with a custom one:
Screen Shot 2021-03-25 at 7 19 34 PM

app/config/schema/rooms.proto
rene-grigory: Good question: I was just following the precedent, but it could certainly be generalized.

hndrewaall:
@rene-grigory do you think you will be able to get this updated to address morg's feedback this weekend? I would really like to get this merged ASAP, as we will need to start working on the new theme(s) soon (and those will depend on any new theme knobs this generates)

rene-grigory:
I think I have addressed all of Morgane's concerns.

app/client/jsx/ChatStreamRoom.jsx
```diff
@@ -30,14 +30,13 @@ export const ChatStreamRoom = ({
...
+          // this._converse.api.listen.on("chatRoomViewInitialized", () => {
+          //   const title = document.querySelector(".chatbox-title__text");
+          //   const dd = document.querySelector(".chatbox-title__buttons");
 
-            title.innerHTML = roomName;
-            dd.remove();
-          });
+          //   title.innerHTML = roomName;
+          //   dd.remove();
+          // });
```
morganecf, line 39: Can we get rid of the comments?

morganecf:
Thanks for getting to all the feedback! I think it looks good -- I would just remove that commented code and also make a ticket for the avatars linking to my initial comment above, so we can get to that later.

rene-grigory:
Oops, good catch.

rene-grigory:
@morganecf & @hndrewaall, I've added a number of styling variables for the Chat and emoji popup. I think everything is ready to be merged in.

hndrewaall:
This looks really, really good. Thank you @rene-grigory ! I will leave to @morganecf for final approval

app/client/styles/themes/_tst.scss
```diff
@@ -13,6 +13,34 @@ $dark-grey: #404040;
...
+$emoji-popup-bg: $matte-purple; // background color of the emoji popup 
+$emoji-separator: $fuckyou-turquoise; // color for th emoji separator bar
+$message-area-bg: $fuckyou-turquoise; // background for the message text area
+$message-input-top-border: $matte-blue; // the border at the top of the message area
+$message-focus: $matte-blue; // the color of the message area border when focused
+$message-text: $matte-blue; // the text color in the message area
+$chat-participants: $fuckyou-turquoise; // the participants sidebar on the right of the chat
+$chat-participants-resize: $matte-blue; // color of the participants resize handle
+$participants-title: $matte-blue; // the title of the participants list
+$participants-name: $matte-blue; // the name of each participant in the list
```
morganecf, line 42: Thanks for adding these comments, that's very helpful

morganecf:
Looks good to me!

PR #451, opened 2021-04-30
Update rooms.json

LEDFlashing:
This might fix the angelthump stream


kacycrider:
I think we can close this PR

Issue #387, opened 2021-04-04
Create Avatars for Chat Room members


rene-grigory:
As part of the review of PR #385, the intent to create Avatars for members in the chat rooms was raised.

We could use the `vcard` API from converse.js to [update the profile information].(https://github.com/conversejs/converse.js/blob/master/src/headless/plugins/vcard.js#L291-L315). The `vcard` will expect a base64 encoded image. The desire is to use the jitsi user's avatar which can be accessed via the generic [Room](https://github.com/morganecf/jitsi-party/blob/master/app/client/jsx/Room.jsx#L82).

[Here is the relevant section from the PR discussing this matter.](https://github.com/morganecf/jitsi-party/pull/385#pullrequestreview-621590867)

PR #488, opened 2021-05-24
Morg/chat avatars

morganecf:
https://github.com/morganecf/jitsi-party/issues/387 

This makes sure the converse chat stream user avatars are the same as the ones in the house. 
Screen Shot 2021-05-24 at 2 06 08 PM'

Note that I set the following CORS policy on the tstvhq bucket so that I could test it from `localhost`. Not sure if we want to get rid of it now? 
Screen Shot 2021-05-24 at 2 08 00 PM


app/client/jsx/JitsiVideo.jsx
```diff
@@ -140,7 +140,7 @@ class JitsiVideo extends Component {
                 this.makeJitsiTransparent()
                 this.hideSpinner()
                 const { type, color } = this.props.jitsiData.avatar
-                const avatarUrl = Config.baseUrl + Config.avatars[type].images[color]
+                const avatarUrl = Config.avatars[type].images[color]
```
morganecf, line 143: This also fixes the avatar display in regular jitsi rooms.

rene-grigory:
We should probably consider allowing the background color to be altered allowing the png to show better. I don't see where Converse.js allows this (it appears to default to a white background). We could override the style like so:
```
.avatar {
  background-color: <whatever color> !important;
}
```

morganecf:
@rene-grigory  this doesn't look amazing but I created a variable for the avatar background color and backfilled it in the latest two themes with the same color we use in the sign in page. 

Screen Shot 2021-05-25 at 4 51 31 PM

hndrewaall (approved):
Thank you so much!

PR #490, opened 2021-05-27
Jukebox

rene-grigory:
# Issues
A request was made to extend the **ImageMapRoom** so that it could present a Jukebox such that when a selection is made the MP3 will open in the Modal dialog used by the PDFs. An additional request was that the room be configured so that HTML pages could be opened as well.

# Solutions
- A new `type` was added to the imagemaps.json config which identifies whether an enpoint is a PDF, MP3, or HTML page. 
- A check was put in place to identify whether we are presenting a list of PDFs or individual MP3s and HTML pages.
- For MP3's the Modal dialog was repurposed to present an MP3 play which auto starts the song.
- For HTML pages, since some of the pages have `x-frame-options: DENY`, we redirect to the link in a new browser tab
- Rooms have been configured both in TSTVHQ and Faction Hall.

# Comments
- I discovered a couple of SASS variables were missing from the _factionhall.scss; so I added them.
- I also discovered an infinite re-render in the **ImageMapRoom**. An empty `img` was being generated and it's onLoad function was set to set a state variable: `setImgSize`. The problem with this is the page loads and the image loads setting the state which causes a re-render which causes the image to re-load, setting the state again, rinse and repeat for eternity. I changed the `setImgSize` temporarily to a `imgSizeRef` which prevents the infinite re-render. However, the way the image is presented needs work so this may not be a permanent solution.


morganecf (approved):
Looks good to me!

Issue #512, opened 2021-06-08
Make map size a configurable css variable


morganecf:
This is to avoid situations like the one in https://github.com/morganecf/jitsi-party/pull/513 where we hardcode the map proportions to fit certain map designs.

rene-grigory:
@morganecf, I'm looking at the code and I see that the **width** and **height** are being passed to the **MapVisualization** component which generates the SVG. The **d3** utility that generates the SVG uses the **width/height** along with the **padding** value to perform a linear scaling. It may be easier to make the height and width ratios config settings instead of replacing everything with CSS.

Something like the following: 
```
{
  mapSizeRatio: {
    width: 2.5,
    height: 1.25
  }
}
```

rene-grigory:
I can take care of it.

PR #513, opened 2021-06-08
Make map more narrow

morganecf:
…for now!

This is a temp fix. Ticket to actually fix this the real CSS way here: https://github.com/morganecf/jitsi-party/issues/512 

image

### DIFF
--- a/app/client/jsx/Map.jsx
+++ b/app/client/jsx/Map.jsx
@@ -26,7 +26,7 @@ class Map extends Component {
                 return room
             })
 
-        const width = document.querySelector('.map').clientWidth / 1.5
+        const width = document.querySelector('.map').clientWidth / 2.5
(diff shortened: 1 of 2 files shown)


app/client/jsx/Map.jsx
```diff
@@ -26,7 +26,7 @@ class Map extends Component {
                 return room
             })
 
-        const width = document.querySelector('.map').clientWidth / 1.5
+        const width = document.querySelector('.map').clientWidth / 2.5
```
morganecf, line 29: There's some autoformatting below but this is the only relevant change

hndrewaall (approved):

rene-grigory (commented):
@hndrewaall & @morganecf , I've looked this over and it looks good to me; if you all want to merge it in.

PR #525, opened 2021-06-15
hack to fix when the chatroom vanishes

rene-grigory:
# Issue
Occasionally a user's chatroom will vanish. The only reason I can figure for this behavior is they are somehow disconnected from Prosody.

# Solution
Unfortunately, Converse.js consumes the errors and logs some output. Therefore, I had to detect when the chatroom actually vanishes using a MutationObserver. When it disappears I need to logout of converse, clear both the local and session storage of any "converse"  related keys and then re-initialize everything. This is a hack, but it seems to work. We will definitely need to monitor going forward.


hndrewaall (approved):
Brilliant. Thank you so much!

PR #537, opened 2021-06-22
added show/hide feature

rene-grigory:
# Issue
A number of folks reached out to me this week complaining that the chat scrolling, while they are watching a service, can be incredibly distracting. 

# Solution
This PR adds a Show/Hide Chat icon in the navigation. The chat will display by default, but when the icon is clicked the chat will hide or display as needed.


hndrewaall:
Nice! I think it would be nice if the video iframe resized to take up in the whole width when the chat is hidden, but this definitely seems to work for now

PR #556, opened 2021-06-30
Basic build automation with Github Actions

hndrewaall:
It works, this time! Probably.


hndrewaall:
@morganecf for this to work on merge, you'll need to set up a Docker Hub account, after which I can grant you push access to the `cabinweekend` org. Then you'll need to create an API token, which you'll need to then add to this repo's secrets.

The two secrets that need to be defined are `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN`

PR #583, opened 2021-07-20
Update events.json

kacycrider:


app/config/temple/events.json
```diff
@@ -1,51 +1,43 @@
 {
     "events": [
         {
-            "name": "Knowing Our History: Constructing Identity & Building Pride. Led by: Chalice Blythe",
+            "name": "Topic: The Power of Naming Yourself. Led by: Minister Hadrian Flyte,
```
hndrewaall, line 4: You dropped a quote here. interesting that the checks didn't catch it

hndrewaall (changes requested):

app/config/temple/events.json
hndrewaall: @morganecf how hard would it be to add frontend tests to catch this sort of thing? like, i am pretty sure this would generate a react error

hndrewaall: though i honestly don't understand why it doesn't break at the protoc level. maybe cuz that's only for rooms?...

PR #590, opened 2021-07-23
Update imagemaps.json

kacycrider:


app/config/tstvhq/imagemaps.json
```diff
@@ -77,368 +73,279 @@
                     "contents": [
                         {
                             "title": "What does it remind you of?",
-                            "path": "https://tstvhq.s3.amazonaws.com/imagemap-content/staircase-map/topic/Mesner\u2014Repressed-Memory.pdf",
-                            "type": "PDF"
```
hndrewaall, line 81: I think you need to restore these `type` lines

hndrewaall (changes requested):

app/config/tstvhq/imagemaps.json
kacycrider: Okay thanks-- this image map isn't in use to be clear, but I added the missing lines in case it's breaking something else

hndrewaall: Sorry, you misunderstood: all the lines you deleted that say `"type": "PDF"` need to be restored

hndrewaall (changes requested):

hndrewaall:
Also, you probably need to add that line to all the other objects now

hndrewaall:
will merge and fix locally

PR #591, opened 2021-07-23
Update _hexennacht.scss

kacycrider:


hndrewaall:
wait, we are using the hexennacht theme? why?

kacycrider:
The Hexennacht theme was the last one implemented at VHQ. 
I want the Temple CSS to apply to VHQ. 
If you would like me to copy the Temple theme into another CSS I can-- Let me know what CSS you would like me to copy the Temple CSS into so that it applies to VHQ.

hndrewaall:
Doing it that way is very confusing. Let's keep the titles of themes meaningful. If it's called hexennacht, leave it for hexennacht.

I think you have three reasonable options:

1. Update the tst css. This follows the logic of it being a generic theme that we update at will.
2. Create a new theme called unveiling or whatever. Put the contents of the temple theme here. Then we need to update the VHQ build to point to this theme, but that's easy.
3. Just point our VHQ build at the temple theme. This is just the second part in step 2.

kacycrider:
Thank you! That's very helpful. 

I've copied the Temple CSS into TST CSS as per the first option.

hndrewaall:
Er, no you haven't

hndrewaall:
But also I was editing that comment

hndrewaall:
Actually all three require the same change. Your option here, hexennacht, is indeed the fewest changes, but it's wrong conceptually for the reason I mentioned.

Here's the place you can change which theme the job uses: https://github.com/morganecf/jitsi-party/blob/master/.github/workflows/build-test-push.yml#L34

hndrewaall:
Please don't open new PRs, just update this one. When you make new PRs in response to my comments here it is very hard to follow.

hndrewaall:
I guess it is not very obvious how to do that from the web ui. But, here's a link that should help: https://github.com/kacycrider/jitsi-party/tree/patch-248

If you edit files from that view you should be able to commit to this same branch, which means the changes will show up here.

You can find that link up here: 
image

kacycrider:
I think I did what you asked please let me know what else I need to do. Thank you

hndrewaall:
You have not. You haven't made any changes to this PR since my comment. You opened a new PR, with only one of the two changes needed. 

You need to update the target theme (as you did in the other PR) but you ALSO need to actually update the theme you are pointing it to (in the case of your other PR, `tst`).

If you can't figure out how to edit this PR right now, that's okay. But if so you will need to open a third PR with the changes to the theme I just mentioned. After which we can merge those two and I will close this one without merging.

PR #592, opened 2021-07-23
Update _tst.scss

kacycrider:


kacycrider:
This is my attempt to follow step 1 from the comment in hexennacht CSS -- copying temple into TST CSS

hndrewaall:
Sorry, this one never showed for me. Cool, so you did do both parts. This is the problem with opening multiple PRs instead of updating a single one: they are hard to keep track of.

hndrewaall:
Why did you close this one?

hndrewaall:
If you reopen this one, we can merge this and the other one and close the first.

